PR #78, opened 2026-05-03
fix(publisher-ingest): retract events when a publisher is disabled

bbernstein:
## Summary

- Disabling a publisher (`enabled: false`) previously stopped *fetching* their feed but left their already-published events live on the sidecar JSON. This breaks the moderation use case — disable a misbehaving publisher → their content stays live.
- Adds a disable-retraction loop in `runIngest` that runs after the enabled-publisher loop and before the sidecar republish: every disabled publisher's events are hard-deleted via batched TransactWrite (chunks of 100). Per-publisher errors are logged and swallowed so one bad publisher cannot block the rest of the run or the sidecar publish.
- Hard-delete bypasses the reconciler on purpose — the reconciler skips past-event removals and would trip the threshold halt on a 100% removal.

## Changes

- `PublisherRegistryService.listDisabled()` — mirror of `listEnabled()`, filters `enabled = false`.
- `PublisherEventStore.deleteAllForPublisher(id)` — lists all events for a publisher and hard-deletes them in batched TransactWrite. Returns the deleted count.
- `runIngest` — disable-retraction loop after the enabled-publisher loop, before `listAllPublished` + sidecar publish.
- 6 new tests (3 unit, 3 integration). Existing integration tests updated to declare the new mock methods.
- Plan doc at `docs/plans/2026-05-03-disable-publisher-retracts-events-plan.md` (scoping commit + implementation-status footer noting the small deviations from plan).

## Deviations from the plan (called out in the plan doc footer)

- Used a new `listDisabled()` method instead of `listAll().filter(p => !p.enabled)` — parallel structure with `listEnabled()`, avoids fetching enabled rows we'd discard.
- Deferred the optional `recordFetchOutcome(p.id, { status: 'ok', message: 'disabled — retracted N event(s)' })` write (and the `disabled` `FetchStatus` enum the plan flagged as an open question). Easy follow-up if the admin UI signal is wanted.

## Test plan

- [x] `cd backend && npx jest --testPathPattern='publisher(EventStore|RegistryService|IngestHandler)'` — 35 tests pass.
- [x] `cd backend && npx jest` — full backend suite, 240 tests pass.
- [x] `cd frontend && npm run build` — passes.
- [ ] Live verification after deploy: `bbtest` publisher is currently `enabled: false` with 2 events still on the sidecar. After deploy + one ingest tick, both events should disappear from `chautauqua-calendar-publisher-events` and the sidecar JSON with no manual action.

### DIFF
--- a/backend/src/__tests__/publisherEventStore.test.ts
+++ b/backend/src/__tests__/publisherEventStore.test.ts
@@ -64,6 +64,45 @@ describe('PublisherEventStore', () => {
     expect(mockSend).toHaveBeenCalledTimes(2);
   });
 
+  it('deleteAllForPublisher is a no-op when no events exist', async () => {
+    mockSend.mockResolvedValue({ Items: [] });
+    const n = await store.deleteAllForPublisher('p');
+    expect(n).toBe(0);
+    // Only the list query, no TransactWrite
+    expect(mockSend).toHaveBeenCalledTimes(1);
+  });
+
+  it('deleteAllForPublisher issues a single TransactWrite with all event keys', async () => {
+    mockSend
+      .mockResolvedValueOnce({
+        Items: [
+          { publisherId: 'p', eventId: 'e1', state: 'published' },
+          { publisherId: 'p', eventId: 'e2', state: 'published' },
+        ],
+      })
+      .mockResolvedValueOnce({});
+    const n = await store.deleteAllForPublisher('p');
+    expect(n).toBe(2);
+    expect(mockSend).toHaveBeenCalledTimes(2);
+    const txn: any = mockSend.mock.calls[1][0];
+    expect(txn.input.TransactItems).toHaveLength(2);
+    expect(txn.input.TransactItems[0].Delete.Key).toEqual({ publisherId: 'p', eventId: 'e1' });
+    expect(txn.input.TransactItems[1].Delete.Key).toEqual({ publisherId: 'p', eventId: 'e2' });
+  });
+
+  it('deleteAllForPublisher chunks at 100 keys per TransactWrite', async () => {
+    const items = Array.from({ length: 250 }, (_, i) => ({
+      publisherId: 'p', eventId: `e${i}`, state: 'published',
+    }));
+    mockSend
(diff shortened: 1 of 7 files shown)


bbernstein:
Addressed the claude-review feedback in 0686ffc:

- **Issue 2 (perf)**: replaced `listEnabled()` + `listDisabled()` with a single `listAll()` call and an in-memory split. DynamoDB Scan reads every row regardless of FilterExpression so two scans was wasteful, as you noted.
- **Issue 3 (missing `enabled` attribute)**: implicitly fixed by the refactor — `!p.enabled` treats undefined as falsy, so missing-attribute rows now fall into the disabled bucket and get retracted. Added a code comment in `runIngest` calling this out so it doesn't regress.
- **Observation (silent retraction count)**: now logs `[publisher-ingest] retracted N event(s) for disabled publisher <id>` to CloudWatch when N > 0.
- **Removed `listDisabled()`** — it was added in the prior commit but became dead code after the refactor.
- **Issue 1 (unguarded list call)**: implicitly resolved. There's no longer a separate disabled-publisher Scan that could fail independently — `listAll()` is the single failure point and its risk profile matches the pre-PR behavior of `listEnabled()`. Did not add a separate try/catch since there's no new failure surface beyond what already existed.

Re-requesting review.

bbernstein:
All review comments have been addressed.

- **Round-1 claude-review** (3 issues + 1 observation): all resolved in 0686ffc — single `listAll()` Scan with in-memory split, retraction count logged when N > 0, missing-`enabled`-attribute case now handled implicitly via `!p.enabled`.
- **Round-2 claude-review**: passed. Quote: *"Looks good to merge."* One non-blocker observation (existing `PublisherRegistryService.listEnabled()` is now production-dead, but it's pre-PR code so cleanup is out of scope here). No threads need manual resolution.
- **Checks**: all 5 SUCCESS — `test-backend (20.19.0)`, `test-backend (22)`, `test-frontend (20.19.0)`, `test-frontend (22)`, `claude-review`.
- **Mergeable**: `MERGEABLE` / `mergeStateStatus: CLEAN`.

PR is ready for your merge.

PR #95, opened 2026-05-04
feat(admin): surface detailed fetch errors on /admin/publishers/

bbernstein:
## Summary
- The /admin/publishers/ "Last Fetch" column now shows the timestamp and, on non-ok status, the full `lastFetchMessage` inline (previously only visible as a hidden tooltip on the badge).
- Backend stores a more diagnostic message: error paths are prefixed (e.g. `/publisher/id: ...`), and `err.cause` is unwrapped so admins see `ENOTFOUND host` / `ECONNREFUSED` / `timed out after 30000ms` instead of a bare "fetch failed".
- Validation `mismatch` errors now spell out both the feed and registered publisher ids.

## Test plan
- [x] `frontend && npm run validate && npm run build`
- [x] `backend && npm test` (420 tests pass)
- [x] `frontend && npm test` (130 tests pass)
- [ ] After deploy, verify a failing publisher row on /admin/publishers/ shows the detailed message inline (not just in tooltip)

### DIFF
--- a/backend/src/__tests__/publisherFeedFetcher.test.ts
+++ b/backend/src/__tests__/publisherFeedFetcher.test.ts
@@ -73,6 +73,7 @@ describe('fetchAndParseFeed', () => {
     expect(r.fetchStatus).toBe('validation_error');
     expect(r.feed).toBeNull();
     expect(r.report.errors[0].message).toContain('different-id');
+    expect(r.report.errors[0].message).toContain('test-pub');
   });
 
   it('returns network_error when fetch throws', async () => {
--- a/backend/src/__tests__/publisherFeedFetcher.test.ts
+++ b/backend/src/__tests__/publisherFeedFetcher.test.ts
@@ -57,6 +57,22 @@ describe('fetchAndParseFeed', () => {
     );
     expect(r.fetchStatus).toBe('validation_error');
     expect(r.feed).toBeNull();
+    // Mismatch message must spell out both ids — admins triage by reading
+    // this string in the publishers table and need to know what the feed
+    // claims its id is vs. what we have registered.
+    const msg = r.report.errors[0].message;
+    expect(msg).toContain('different-id');
+    expect(msg).toMatch(/test-pub|publisher.id/);
+  });
+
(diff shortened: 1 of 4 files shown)


bbernstein:
Thanks for the thorough review. All four issues addressed in 8ebbd55:

**#1 Cause duplication** — switched to `err.cause.message || err.cause.code` in `publisherFeedFetcher.ts`. Added a regression test asserting `ENOTFOUND` appears exactly once in the surfaced message.

**#2 Misplaced cause unwrapping in handler** — agreed; the outer catch in `publisherIngestHandler.ts` fires for any unhandled throw (DDB, reconcile, etc.), not network errors. Reverted that block to the simple form. `fetchAndParseFeed` keeps the cause unwrapping where it's actually meaningful.

**#3 AbortError + cause edge case** — added an early return for `err.name === 'AbortError'` so the deterministic timeout message isn't decorated with a DOMException cause. New test verifies the message stays clean (`"timed out after 30000ms"`, no leaked "aborted").

**#4 Test coverage** — added four new tests in `publisherFeedFetcher.test.ts`:
- AbortError → clean timeout message, no cause leak
- `err.cause` with both message+code → no duplication, message surfaced
- `err.cause` with code only → falls back to code
- HTML mismatch path → asserts both ids appear in the error
- Updated the existing JSON mismatch test to assert both ids appear

Backend suite: 424 tests pass (4 new). Frontend unchanged.

@claude please re-review.

bbernstein:
All review comments have been addressed.

**Verification:**
- ✅ All 5 CI checks passing: `test-backend (24)`, `test-backend (25)`, `test-frontend (24)`, `test-frontend (25)`, `claude-review`
- ✅ Two `claude-review` runs (initial + re-review) both returned LGTM
- ✅ No pending reviewers
- ✅ `mergeStateStatus`: CLEAN, `mergeable`: MERGEABLE
- ✅ Backend suite: 424 tests pass; frontend suite: 130 tests pass

**No threads need manual resolution** (review feedback came as inline issue comments, not as review threads).

The PR is ready for merge — please merge when convenient.

PR #101, opened 2026-05-06
fix(sw): evict cached entry on non-OK response (retracted publisher events)

bbernstein:
## Summary

The publisher-disable retraction path is correct end-to-end on the backend (PR #78), but the service worker's `staleWhileRevalidate` handler caches the publisher-events sidecar indefinitely once it's seen the file 200 OK once: when the sidecar later 404s (because the only event was retracted, or the year now has zero published events and `PublisherSidecarPublisher.publish()` deletes the file), the background revalidate gets a 404, the handler no-ops on non-`ok` responses, and the stale cached body keeps being served.

Reproduced in prod with `pub-7e7c2377-…`:
- `chautauqua-calendar-publishers` row: `enabled=false` ✓
- `chautauqua-calendar-publisher-events` query for that publisherId: 0 rows
- `s3://…/cache/calendar-cache/publisher-events-2026.json`: 404
- `all-events-2026.json`: no trace of the publisher
- Browser: still rendering the retracted event because `chqcal-v2-data` Cache Storage holds the old body

## Change

Two-line fix in `frontend/public/sw.js`:
- `staleWhileRevalidate`: on non-OK responses, `cache.delete(request)` so the next load sees no cached entry and the authoritative 404 propagates through. The `.catch` branch (network errors) still no-ops so transient offline blips don't wipe valid cached bodies.
- Bumped `CACHE_VERSION` `chqcal-v2` → `chqcal-v3` so the activate handler's old-cache sweep flushes affected clients' caches on next load (without the bump they'd need one extra reload after the SW update for the new eviction logic to clear the lingering entry).

## Test plan
- [x] `npm run validate` (typecheck + lint)
- [x] `npm run build`
- [x] `npx vitest run` (207 tests pass — no SW-specific tests; SW lives in `public/` and isn't unit-testable in the current vitest setup)
- [ ] Manual after deploy:
  - DevTools → Application → Cache Storage shows `chqcal-v3-data` (not v2) on next visit
  - Disable a review-mode publisher with an approved event → run "Ingest Now" → reload → event no longer appears
  - With DevTools Network tab open, the sidecar 404 visibly arrives and the cache entry disappears (verifiable via Application → Cache Storage)

### DIFF
--- a/frontend/public/sw.js
+++ b/frontend/public/sw.js
@@ -1,4 +1,4 @@
-const CACHE_VERSION = 'chqcal-v2';
+const CACHE_VERSION = 'chqcal-v3';
 const STATIC_CACHE = CACHE_VERSION + '-static';
 const DATA_CACHE = CACHE_VERSION + '-data';
 
@@ -100,10 +100,19 @@ async function staleWhileRevalidate(event, cacheName) {
   const cache = await caches.open(cacheName);
   const cached = await cache.match(request);
 
-  // Fetch fresh data in background regardless
+  // Fetch fresh data in background regardless. On a non-OK response (e.g.
+  // a sidecar that was deleted upstream after a publisher was disabled or
+  // had its last event retracted) we must EVICT the cached entry — leaving
+  // it behind would let stale-while-revalidate serve the deleted payload
+  // forever, since this branch only ever runs on subsequent loads and the
+  // 404 itself wouldn't otherwise dislodge the prior good cache. Network
+  // errors (the .catch branch) intentionally do nothing so a transient
+  // offline blip doesn't wipe a still-valid cached body.
   const fetchPromise = fetch(request).then((response) => {
     if (response.ok) {
       cache.put(request, response.clone());
+    } else {
+      cache.delete(request);
     }
     return response;
   }).catch(() => null);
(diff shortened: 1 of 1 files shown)


bbernstein:
Addressed claude-review's feedback in 344010c: scoped the eviction to 404 only.

> Edge case worth discussing: evicting on 5xx, not just 404 [...] A transient 500/503 from CloudFront during the background revalidation would also evict a perfectly valid cached entry.

Agreed — 404 is the only status that actually signals "the resource was deleted upstream" (which is what `PublisherSidecarPublisher.publish()` produces via `DeleteObject` when a year has zero published events). 403/429/5xx are transient or auth signals, not deletion signals; preserving the cached body in those cases is the right call so a momentary origin hiccup doesn't take down the calendar.

Updated the comment in `sw.js` to spell out the 404-only scope and rationale.

bbernstein:
## Ready for merge

All review comments have been addressed:

- ✅ claude-review (round 1) — flagged 5xx eviction concern. Addressed in 344010c by scoping `cache.delete` to `response.status === 404` and documenting the rationale in the SW comment block.
- ✅ claude-review (round 2) — approved. _"The 404-only scope is the right call. The change is complete and correct. No further issues."_

No threads need manual resolution (review comments were issue-level, not inline review threads).

All checks passing:
- test-frontend (Node 24, 25): SUCCESS
- test-backend (Node 24, 25): SUCCESS
- claude-review: SUCCESS

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

PR is ready for merge.

PR #124, opened 2026-05-12
chore(backend): set up ESLint + clear all warnings + wire into CI

bbernstein:
## Summary

Brings the backend up to the same lint/type discipline the frontend already
enforces, and wires it into CI so regressions can't ship.

**Per [`docs/plans/2026-05-12-backend-lint-and-cleanup-plan.md`][plan]:**

- **Phase 1** (`a633a3b`): adds eslint flat-config matching frontend major
  versions (eslint 9 / typescript-eslint 8), plus `lint` / `type-check` /
  `validate` scripts. Baseline captured: 26 warnings, 0 errors — all
  `@typescript-eslint/no-unused-vars` in production code.
- **Phase 2** (`b3cb363`): clears all 26 warnings (deletes dead imports,
  underscore-prefixes unused Lambda handler params, drops empty catch
  bindings) and fixes one latent tsc error in `syncHandler.ts` that
  esbuild had been silently ignoring. Per user request, the lint script
  now runs with `--max-warnings=0` so any future warning fails the build.
- **Phase 3** (`d7fd08c`): adds `npm run validate --workspace=backend`
  as a CI step in the `test-backend` job, and updates `CLAUDE.md`'s
  verification checklist to include the backend equivalent.

Phase 4 (opportunistic dead-code / comment scan) intentionally deferred —
the plan marks it skippable and these changes are already a coherent
unit.

[plan]: ../blob/main/docs/plans/2026-05-12-backend-lint-and-cleanup-plan.md

## Test plan

- [x] `npm run validate --workspace=chautauqua-backend` exits 0 locally
- [x] `npm test --workspace=chautauqua-backend` — 808/808 pass locally
- [ ] CI build-and-test workflow green on this PR (new backend-validate
      step exercises the lint gate)
- [ ] Sanity: confirm CI fails if a warning is intentionally introduced
      (per plan's Self-review checklist) — can be done in a throwaway
      commit on this branch before merging if reviewers want belt-and-suspenders


bbernstein:
Pushed `9d26591` addressing all four review threads:

**Copilot inline #1** (`calendarHandler.ts` date extraction) — fixed by switching to `substring(0, 10)`, which extracts the date from both DB-format (`YYYY-MM-DD HH:MM:SS`) and ISO-format (`YYYY-MM-DDTHH:mm:ss…`) strings consistently. The previous `split(' ')[0]` worked for DB strings but returned the entire ISO string (with time) for ISO inputs, breaking single-day filters. The dead `Date()` parses removed earlier in this PR had been masking how broken that comparison was for ISO values. Resolves the latent bug Copilot flagged.

**Copilot inline #2** (`syncHandler.ts` cast bypasses validation) — added a runtime whitelist (`VALID_SYNC_TYPES`) and return 400 with the allowed set when the query-string `type` doesn't match. The TS cast still narrows the union after validation passes.

**Copilot inline #3** (`CLAUDE.md` `--max-warnings=0` claim) — reworded to state explicitly that only the backend lint script uses the flag; frontend warnings are reported but don't fail.

**Claude review note** (extra blank line in `eventTransformationService.ts` after removing `endDate`) — fixed.

Three Copilot threads should be marked resolved once the bot agrees the fixes match its concerns. Re-requested Copilot review below.

Verification: `npm run validate --workspace=chautauqua-backend` exits 0; full test suite still 808/808.

bbernstein:
Pushed `62e3c35` addressing Copilot's second review:

**Copilot #1 (`limit` validation)** — added bounds check in `syncListHandler`: integer, 1–100, otherwise 400. Same shape as the `type` validation. Mirrors the gap Copilot identified.

**Copilot #2 (`VALID_SYNC_TYPES` duplicates `SyncStatusRecord['type']`)** — moved the const to `syncStatusService.ts` and made the array canonical:

```ts
export const VALID_SYNC_TYPES = ['manual', 'scheduled', 'full', 'incremental', 'daily', 'hourly'] as const;
export type SyncType = typeof VALID_SYNC_TYPES[number];

export interface SyncStatusRecord {
  type: SyncType;
  // …
}
```

The handler now imports both. Drift is impossible because the type is derived from the const.

**Copilot #3 (unit tests for date-range filter)** — intentionally deferred. The filter is embedded inside `queryEventsFromDatabase` in `calendarHandler.ts`, so testing it cleanly requires extracting it. That's scope creep for a lint-cleanup PR. I'll track it as a follow-up issue/PR.

Threads should be marked resolved once Copilot agrees with the fixes. Re-requesting Copilot review.

Verification: `npm run validate --workspace=chautauqua-backend` exits 0; `npm test --workspace=chautauqua-backend` still 808/808.

bbernstein:
Pushed `7aa8728` addressing the third Copilot review:

**Copilot inline (eslint.config.mjs:46) — test override is dead weight.** Verified: Phase 1B's baseline had 0/26 hits in `src/__tests__/`, so the override was never needed. Removed it; `npm run lint --workspace=chautauqua-backend` still exits 0, so tests now follow the same unused-vars discipline as production code.

**Copilot inline (CLAUDE.md:285) — redundant `validate && build`.** Correct — frontend `build` already calls `validate` (and tests). Replaced with just `npm run build`.

**Copilot inline (syncHandler.ts:249, limit validation).** This concern was already addressed by `7211577` (the `/^\d+$/` regex rejects `"10abc"` and `"1e2"` before `parseInt` runs). Copilot's review must have raced with the push. No new fix needed.

**Copilot inline (syncHandler.ts:254, add tests for the new 400 paths).** Same response as before: handler-level test additions are scope creep for a lint-cleanup PR. I'll track as a follow-up.

**Pre-existing issues** Claude surfaced (`getRecentSyncStatuses` silent fallback to `manual`-only queries, `cleanupOldRecords` using `UpdateCommand` with `REMOVE #id` on the partition key, duplicate `POST /feedback` block in `calendarHandler`): noted, out of scope, will track as separate issues.

Verification: `npm run validate --workspace=chautauqua-backend` exits 0; full test suite still 808/808.

PR should be ready for merge after CI on `7aa8728` goes green. The `test-backend` job will run on the push event for this branch (not the same-repo PR event), so the lint gate is exercised post-push.

bbernstein:
## All review comments addressed — ready for merge

### Final state

- **mergeStateStatus**: `CLEAN`
- **mergeable**: `MERGEABLE`
- **Pending reviewers**: none (`requested_reviewers` empty)
- **CI**: all checks SUCCESS on \`ad8c6a8\` (push event); same-repo PR-event runs are intentionally SKIPPED per the existing job conditional — this is unchanged by this PR and noted by both reviewers
- **Reviews**: Claude approved (`16:07:43Z` review on \`ad8c6a8\`); Copilot's 4 latest inline comments are non-blocking nits — see below

### Copilot's remaining inline comments on \`ad8c6a8\` (deferred with rationale)

1. **Type-guard helper for \`SyncType\`** (`syncHandler.ts:231, 253`) — would be cleaner than the inline \`as SyncType\` cast inside \`Array.includes\`, but Claude noted: *"the cast inside \`includes\` is the standard TypeScript workaround when the checked value isn't already narrowed to the union… A \`isSyncType\` predicate would be cleaner for reuse across multiple handlers, but is pure ergonomics for a single call site."* No regression; runtime validation precedes the cast. Worth refactoring if a second handler ever needs the same check.

2. **PR description workspace name** (`build-and-test.yml:42`) — Copilot flagged that the PR description says \`--workspace=chautauqua-backend\` while the workflow uses \`--workspace=backend\`. **Both forms work** — \`backend/package.json\` has \`"name": "chautauqua-backend"\` and npm accepts either the package name or the path-based selector. No fix needed.

3. **\`initTables.ts:21\` bare \`catch {}\`** — Copilot and Claude both noted that distinguishing \`ResourceNotFoundException\` from other errors would be safer (a transient auth error currently looks like "table doesn't exist"). The bare \`catch\` is unchanged in *runtime* behaviour from the pre-PR \`catch (error)\` — this PR only dropped the unused binding to clear lint. The semantic improvement is a real follow-up but out of scope for a lint-cleanup PR. Will file separately.

4. **Add unit tests for new 400 paths** (\`syncHandler.ts:254\`) — same scope-creep response as prior rounds; the handler now has well-defined 400 contracts so a future test PR is straightforward.

### Pre-existing issues surfaced during review (tracked for follow-up, not blocking)

- \`syncStatusService.getRecentSyncStatuses\` silently falls back to \`type='manual'\` when called without a \`type\` filter (callers expecting all-types get incomplete data)
- \`syncStatusService.cleanupOldRecords\` uses \`UpdateCommand\` + \`REMOVE #id\` on the partition key — should be \`DeleteCommand\`
- Duplicate (unreachable) \`POST /feedback\` block in \`calendarHandler.ts:597\`
- No type-aware ESLint (\`parserOptions.project\` not set) — blocks \`no-floating-promises\` / \`await-thenable\` for future enablement

### Review threads that should be marked resolved

All Copilot threads from R1–R3 (six threads) plus Claude's R1 cosmetic note should be resolved — each was addressed in a specific commit referenced in the comment trail. The four R4 inline comments above are also addressable as resolved (intentional deferrals with rationale).

### Summary

Nine commits over five review rounds. Lint gate is in place; backend now matches frontend's discipline plus a stricter \`--max-warnings=0\` posture. Awaiting your merge.

bbernstein:
Pushed `687ad6c` fixing the three pre-existing bugs that surfaced during the review (per your direction to fix them in this PR before continuing).

### Bug 1 — `getRecentSyncStatuses` no-type fallback

Was silently returning only `manual`-type records when called without a \`type\` filter. Three call sites (\`getSyncStatistics\`, \`cleanupOldRecords\`, \`syncListHandler\` with no filter) all expected all-type data, so this masked records of every other type. Fixed by fanning out one TypeIndex Query per \`VALID_SYNC_TYPES\` entry via \`Promise.all\`, merging, sorting by \`timestamp\` DESC, and slicing to \`limit\`. The table has no "all types sorted by timestamp" GSI so this is the cheapest correct path without adding an index — worst-case work is bounded at 6 × limit.

### Bug 2 — `cleanupOldRecords` used `UpdateCommand` instead of `DeleteCommand`

The cleanup loop sent \`UpdateCommand\` with \`UpdateExpression: 'REMOVE #id'\` on the partition key. DynamoDB rejects (or silently no-ops) attempts to remove a primary key attribute, so old records were never actually deleted. Replaced with \`DeleteCommand({TableName, Key: {id}})\`. Function is only called from TODO-stub code today, which is why the bug never surfaced in production.

### Bug 3 — duplicate `POST /feedback` block

Two identical blocks at \`calendarHandler.ts:346\` and \`:597\`. Second was unreachable. The second did set \`archived: false\` on new records, which was the only behavioral difference — moved that field into the live block, deleted the dead one.

### Test updates

The \`getRecentSyncStatuses\` no-type fix changes mockSend's call shape (was 1 call, now 6), so the corresponding tests needed to be refactored:
- "without type filter" now asserts call count = `VALID_SYNC_TYPES.length`
- **New test** verifies the merge → sort-by-timestamp-desc → slice path with one fixture per type
- `cleanupOldRecords` tests inspect `DeleteCommand.mock.calls` because \`setup.ts\` auto-mocks \`@aws-sdk/lib-dynamodb\` — the constructor is a jest stub so \`instance.input\` isn't preserved, but the constructor's call args are captured
- \`getSyncStatistics\` tests distribute their fixture records across per-type queries instead of returning the same list for every call

Verification: \`npm run validate --workspace=chautauqua-backend\` exits 0; \`npm test --workspace=chautauqua-backend\` — 809/809 pass (+1 new merge/sort test).

The remaining item from the prior "pre-existing issues" list — **No type-aware ESLint** (\`parserOptions.project\` not set) — is a feature enhancement rather than a bug fix, and adding it now would expand scope further. Keeping it as a future follow-up.

bbernstein:
Pushed `0dfb624` addressing the latest round:

**Claude review (pre-existing `getActiveSyncs` bug)** — same shape as the bug 1 fix in \`687ad6c\`: \`getActiveSyncs\` only queried \`type='manual'\`, hiding any in-flight scheduled/full/incremental/daily/hourly sync. Applied the same fan-out pattern (one TypeIndex Query per \`VALID_SYNC_TYPES\` entry → merge → sort timestamp DESC → slice to the 50-row cap). Tests updated to verify the merged ordering.

**Copilot inline (`backend/package.json:19`, prevalidate hook)** — on a fresh checkout, \`npm run validate --workspace=chautauqua-backend\` fails because tsc resolves \`@chq-calendar/publisher-format\` types from \`tools/publisher-format/dist/index.d.ts\` which doesn't exist until publisher-format is built. CI already builds it as a separate step before validate, so this only impacts local dev. Added \`prevalidate: "npm run build:publisher-refs"\` so the documented \"run validate before committing\" workflow works from clean state. Cost is small (publisher-format is a tiny package).

Verification: \`npm run validate --workspace=chautauqua-backend\` exits 0; \`npm test --workspace=chautauqua-backend\` — 809/809 pass.

bbernstein:
## Final state — ready for merge

### Latest CI on `0dfb624`
- All checks SUCCESS or SKIPPED (skipped on same-repo PR-event jobs by design)
- mergeable: \`MERGEABLE\`
- Claude approved this iteration

### Two new Copilot inline comments — deferred with rationale

Both touch pre-existing DynamoDB query semantics in functions that turn out to have **zero production callers**. \`grep -rn ... --include='*.ts' | grep -v __tests__\` confirms only the function definitions themselves exist outside tests:

1. **\`getActiveSyncs\` — `Limit` + `FilterExpression` semantics** (`syncStatusService.ts:303`). Copilot is correct that DynamoDB applies \`Limit\` before the filter, so an active sync sitting beyond position 50 in the index could be missed. The proper fix is `LastEvaluatedKey` pagination. Claude already noted this is harmless at current scale; combined with the fact that no production code calls \`getActiveSyncs\`, the practical risk is zero. Worth a follow-up if/when the function gets wired up.

2. **\`cleanupOldRecords\` — 1000-record-per-type cap** (`syncStatusService.ts:372`). Copilot is correct that calling \`getRecentSyncStatuses(undefined, 1000)\` and filtering by cutoff misses any expired records past position 1000 per type. The proper fix is to query each type with \`KeyConditionExpression: 'type = :type AND timestamp < :cutoff'\` and paginate. Same situation: \`cleanupOldRecords\` has no production caller (Claude flagged this two rounds ago), so the cap doesn't currently lose data.

Both are recommended follow-ups but out of scope for a lint-cleanup PR that's already grown to 11 commits and four bug-fix patches.

### Summary of what landed in this PR

Original scope (Phases 1–3): backend ESLint flat-config matching the frontend's major versions, all 26 baseline warnings cleared, \`--max-warnings=0\` posture, CI gate via \`npm run validate --workspace=backend\`, plus the test override removed once tests proved clean.

Bug fixes added during review (in the spirit of "while you're already in the file"):
- \`calendarHandler\` date-range filter now uses \`substring(0, 10)\` for both DB and ISO formats (Copilot R1)
- \`syncListHandler\` validates \`type\` and \`limit\` query params with 400 + clear allowed-set/range message (Copilot R1, R2, R3, Claude R2)
- \`VALID_SYNC_TYPES\` is the single source of truth for both runtime and type (Copilot R2 + Claude R1 \`satisfies\`)
- \`syncStatusService.getRecentSyncStatuses\` no-type path fans out across all types instead of silently returning manual-only (Claude R3)
- \`syncStatusService.getActiveSyncs\` same fan-out fix (Claude R4)
- \`syncStatusService.cleanupOldRecords\` uses \`DeleteCommand\` instead of \`UpdateCommand\` with \`REMOVE #id\` on the partition key (Claude R3)
- \`calendarHandler\` duplicate POST /feedback block (the unreachable one) removed; \`archived: false\` migrated to the live block (Claude R4)
- \`prevalidate\` npm hook builds publisher-format before tsc so fresh checkouts pass validate (Copilot R5)
- Latent tsc error in \`syncHandler.ts\` fixed (esbuild had been silently swallowing it)

Plus tests for every new behavior (\`syncStatusService.test.ts\` is up to 25 cases across these methods).

### All review threads — final tally

13 specific items addressed across 6 review rounds; 2 deferred with rationale above. \`mergeStateStatus: MERGEABLE\`, no pending reviewers, no failing checks.

**Awaiting your merge.**

bbernstein:
Pushed `481a021` — **net -998 lines** of dead-code removal per the user direction to "delete unused code rather than fix issues in dead paths".

### What was deleted

**Whole files:**
- \`backend/src/services/syncScheduler.ts\` (182 lines) — never imported; scheduling is done via AWS EventBridge cron, not this in-process scheduler
- \`backend/src/__tests__/syncScheduler.test.ts\` (374 lines) — only exercised the deleted file

**Methods (all with zero production callers; tests existed only to test them):**

\`syncStatusService.ts\`:
- \`getActiveSyncs\` (Copilot's R6 \`Limit\`+\`FilterExpression\` concern goes with it)
- \`cleanupOldRecords\` (Copilot's R6 1000-record-cap concern goes with it)
- \`getSyncStatistics\` (wrapped getRecentSyncStatuses)

\`eventsCalendarDataSyncService.ts\`:
- \`cleanupOldEvents\` — TODO-stub that always logged + returned 0; caller now also gone, \`eventsDeleted\` stays at its 0 initial value
- \`getSyncStatistics\` — TODO-stub returning hard-coded empty shape
- \`bulkProcessEvents\` lost its unused \`_result\` param (5 call sites updated)

**Type-level orphans (none imported outside their declaration file):**

\`types/index.ts\`:
- \`FilterDimension\`, \`FilterOption\`, \`EventFilter\`, \`CalendarRequest\`, \`CalendarResponse\` — \`calendarHandler.ts\` has its own local \`CalendarRequest\` that's the actually-used one

\`types/publisher.ts\`:
- \`isApproved\` — duplicate of the live \`isApprovedPublisher\` in \`utils/publisherApproval.ts\`

**Other unused exports:**
- \`export default\` lines from \`eventsCalendarApiClient\`, \`eventsCalendarDataSyncService\`, \`eventTransformationService\` (all using named imports)
- \`currentHarness\` in the integration test harness; the \`_harnessSingleton\` state that backed it was also write-only after the function went, so it's gone too

### Verification

- \`npm run validate --workspace=chautauqua-backend\` exits 0
- \`npm test\` — **773/773 pass** (was 809; -36 tests of dead code)
- \`npm run test:ci\` — backend coverage **84.69%** lines (floor 81.1%, was 83.86%) — coverage went UP because we removed untested code along with its tests

### What this resolves

Both deferred Copilot R6 concerns are now closed by deletion rather than fix:
- \`getActiveSyncs\` Limit/FilterExpression: function is gone
- \`cleanupOldRecords\` 1000-record cap: function is gone

If/when these capabilities are needed in the future, they can be re-added with a correct implementation rather than maintained as broken-but-unused code.

bbernstein:
Pushed `92ed971` addressing all three Copilot R7 inline comments:

**1. \`eslint.config.mjs\` — jest globals scope** ✅ Split into two `files` blocks. Production code now only has `globals.node`; \`globals.jest\` is scoped to `src/__tests__/**`. A stray \`describe(...)\` in a Lambda handler will now trip \`no-undef\`.

**2. \`syncStatusService.getRecentSyncStatuses\` — Promise.allSettled** ✅ Switched from `Promise.all` to `Promise.allSettled`. A single transient AWS failure on one type now logs and produces a partial result instead of 500ing the whole list endpoint.

**3. \`.github/workflows/build-and-test.yml\` — redundant publisher-format build** ✅ Dropped the explicit `Build publisher-format workspace` step since `validate`'s `prevalidate` hook builds it. Down from 3× to 2× per CI job. Added a comment explaining the residual rebuild for future maintainers.

Verification: `npm run validate` exits 0; `npm test` 773/773 pass.

This should be the final iteration unless reviewers find something new. State recap:
- mergeStateStatus: CLEAN, mergeable: MERGEABLE
- All checks SUCCESS / SKIPPED
- Every Copilot/Claude inline thread across 7 review rounds is now addressed (either fixed, deferred-with-rationale, or resolved-by-deletion of the dead code path)
- 14 commits, net -922 lines (added the lint pipeline + bug fixes + tests; deleted ~1000 lines of dead service code)

bbernstein:
## Final state — ready for merge

Latest reviews on \`27c4f05\`:
- **Copilot**: 25/26 files reviewed, **no new comments**
- **Claude**: explicit *"Approve. Ready to merge."*

### Final tally
- mergeStateStatus: \`CLEAN\`
- mergeable: \`MERGEABLE\`
- No pending reviewers
- All CI checks SUCCESS / SKIPPED
- **15 commits** across 8 review rounds; **19 specific review threads** all resolved
- Net diff: lint pipeline + bug fixes + tests, minus ~1000 lines of dead service code that surfaced during review

### Open follow-ups (not for this PR)
- No type-aware ESLint (\`parserOptions.project\` not set) — would unlock \`no-floating-promises\` / \`await-thenable\` in a future pass

**Awaiting your merge.** This was the loop's last actionable item — there is nothing left for me to iterate on.

PR #137, opened 2026-07-16
fix(pwa): auto-update installed home-screen users on every release

bbernstein:
## Problem

Users who installed **chqcal.org** to their iPhone home screen never received new releases — the app stayed frozen on whatever version it first cached.

**Root cause:** the live app shell was served with a year-long immutable cache header:

```
GET https://www.chqcal.org/
cache-control: public, max-age=31536000, immutable
```

That tells every browser (and every iOS home-screen web clip) *"this file never changes for a year — don't re-check."* New deploys landed on S3 and CloudFront was invalidated, but the **device's own cache never asked origin**, so it kept serving the old shell pointing at old hashed bundles.

The bug was an `aws s3 sync` two-pass pitfall in the deploy: pass 1 uploaded HTML with the immutable header, and pass 2's *second* `sync` skipped the byte-identical HTML (sync skips unchanged files), so the intended short-TTL header was never applied.

This also silently defeated the existing hand-written service worker (`public/sw.js`): its network-first `fetch()` for the shell still honored the HTTP cache and got the year-long-cached copy instead of hitting origin.

## What this PR does

1. **Fixes the cache headers (root cause)** in `deploy-production.yml` (CI, the real deploy path) and `scripts/deploy-frontend.sh`. HTML / `manifest.json` / `version.json` / `sw.js` / `data/weekly-themes/*` are now served `no-cache` via `aws s3 cp` (unconditional, unlike the skip-prone `sync`); content-hashed assets stay `immutable`. Because Pass 1 (all hashed bundles) completes fully before Pass 2, and `index.html` is uploaded before `version.json` within Pass 2, a client that observes the new `version.json` is guaranteed the new shell and its referenced bundles are already on origin.
2. **Adds a client version check** (`frontend/src/lib/versionCheck.ts`) — fetches `/version.json` on load and on app-reopen (`visibilitychange`→visible) and **silently reloads** when the deployed version differs. Includes a `sessionStorage` loop-guard (protects against a CloudFront propagation race) and swallows all fetch failures.
3. **Stamps the build version** — bakes `VITE_APP_VERSION` (git short SHA) into the bundle via Vite `define` and emits a matching `version.json` at build time.
4. **Wires the watcher** into the main calendar entry only (`main.tsx`).
5. **Corrects the design doc** — it had wrongly claimed no service worker exists; documents the existing `sw.js` and how the fix repairs its update path too.

## Honest limitation

There's **no HTTP mechanism that instantly reaches a device already trapped** on the year-long immutable copy — it won't phone home until iOS evicts that cache entry on its own (which iOS does regularly, but not on a schedule we control). Those devices recover the first time they revalidate after this ships, and are self-updating from then on. **Every future release updates automatically.**

## Testing

- New module: 11 unit tests (decision matrix, loop guard, all fetch-failure modes, visibility re-check).
- Full frontend suite green: **407 tests** passing.
- Deploy files are CI/AWS-only — verified by YAML parse + `bash -n` + review of AWS CLI semantics.

**Post-deploy verification** (run once live):
```bash
curl -sI https://www.chqcal.org/            | grep -i cache-control   # -> no-cache, NOT immutable
curl -sI https://www.chqcal.org/version.json | grep -i cache-control  # -> no-cache
```

## Review process

Built via brainstorm → spec → plan → subagent-driven execution (6 tasks, each with an independent spec+quality review) → final whole-branch review (verdict: ready to merge). Design + plan committed under `docs/superpowers/`.

## Deferred non-blocking follow-ups

- Retired HTML pages are no longer auto-pruned from the bucket (excluded from the `--delete` pass; orphaned-but-harmless, served `no-cache`).
- `reload()` sits inside the fetch try/catch (only affects injected test doubles; production `location.reload` never throws).

https://claude.ai/code/session_012ZFzBKLppwMXAu8xvfYQvJ

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -410,20 +410,37 @@ jobs:
       - name: Deploy frontend to S3 and CloudFront
         working-directory: ./frontend
         run: |
-          # Sync to S3 bucket
-          aws s3 sync out/ s3://${{ secrets.S3_BUCKET_NAME }}/ \
+          BUCKET="${{ secrets.S3_BUCKET_NAME }}"
+
+          # Pass 1 — content-hashed, immutable assets. Exclude the files that must always
+          # revalidate so they never get the year-long immutable header. --delete removes
+          # stale hashed bundles; excluded files are neither uploaded nor deleted here.
+          aws s3 sync out/ "s3://$BUCKET/" \
             --delete \
             --exclude "*.map" \
             --exclude "cache/*" \
-            --cache-control "public, max-age=31536000, immutable" \
-            --metadata-directive REPLACE
+            --exclude "*.html" \
+            --exclude "manifest.json" \
+            --exclude "version.json" \
+            --cache-control "public, max-age=31536000, immutable"
 
-          # Update HTML files with shorter cache
-          aws s3 sync out/ s3://${{ secrets.S3_BUCKET_NAME }}/ \
+          # Pass 2 — always-revalidate files. `cp` re-uploads unconditionally (unlike
+          # `sync`, which skips unchanged files and was the cause of the stale-shell bug),
+          # so the no-cache header is applied every time.
+          aws s3 cp out/ "s3://$BUCKET/" \
+            --recursive \
             --exclude "*" \
             --include "*.html" \
-            --cache-control "public, max-age=3600" \
-            --metadata-directive REPLACE
+            --content-type "text/html" \
+            --cache-control "no-cache"
+
+          aws s3 cp out/manifest.json "s3://$BUCKET/manifest.json" \
+            --content-type "application/json" \
+            --cache-control "no-cache"
+
+          aws s3 cp out/version.json "s3://$BUCKET/version.json" \
+            --content-type "application/json" \
+            --cache-control "no-cache"
 
           # Invalidate CloudFront cache
           aws cloudfront create-invalidation \
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -422,6 +422,7 @@ jobs:
             --exclude "*.html" \
             --exclude "manifest.json" \
             --exclude "version.json" \
+            --exclude "sw.js" \
             --cache-control "public, max-age=31536000, immutable"
 
           # Pass 2 — always-revalidate files. `cp` re-uploads unconditionally (unlike
@@ -440,6 +441,10 @@ jobs:
 
           aws s3 cp out/version.json "s3://$BUCKET/version.json" \
             --content-type "application/json" \
+            --cache-control "no-cache"
+
+          aws s3 cp out/sw.js "s3://$BUCKET/sw.js" \
+            --content-type "text/javascript" \
             --cache-control "no-cache"
 
           # Invalidate CloudFront cache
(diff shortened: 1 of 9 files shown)


bbernstein:
Thanks for the review — the weekly-themes finding is a real catch. Verified and addressed:

**✅ Fixed: `data/weekly-themes/<year>.json` served immutable (commit c6d10a3)**
Confirmed the exposure: `useWeeklyThemes.ts:41` fetches `/data/weekly-themes/${year}.json` with no dev/prod split, unlike `useArticleLinks.ts:46` and `useAvailableYears.ts:48` which route through `/cache/` in production. So in prod this file is served straight from the S3 build output and, being non-content-hashed, a content edit under the same filename would never revalidate — exactly the bug class this PR fixes. Both deploy paths now exclude `data/weekly-themes/*` from the immutable Pass 1 and upload it `no-cache` in Pass 2. (The other `/data/` files — `all-events*.json`, `article-links-*.json`, `years.json` — are only read from `/data/` in dev; prod reads them via `/cache/`, so their immutable header is functionally inert. Scoped the fix to the one file actually fetched from `/data/` in prod.)

**✅ Fixed: PR-body upload-order wording**
You're right — actual Pass-2 order is `*.html` → `manifest.json` → `version.json` → `sw.js` (→ now `weekly-themes`). Reworded the PR body to state the invariant that actually matters: Pass 1 (hashed bundles) fully precedes Pass 2, and `index.html` is uploaded before `version.json`, so a client observing the new `version.json` always finds the new shell + bundles already on origin.

**Deferred (non-blocking), with reasons:**
- **`*.xml`/`*.txt` special-case dropped** — no such files exist in the current build output (no `robots.txt`/`sitemap.xml`), so it's a no-op today. Adding one later would need an exclude; not worth a speculative rule now. Noted for whoever adds one.
- **`sessionStorage.setItem` silent no-op → reload loop** — agree it's a real edge but very low-probability (setItem failures typically throw `QuotaExceededError`, already caught → no reload). Since silent no-op (vs. throw) is rare and you flagged it for awareness only, leaving as-is rather than adding an in-memory guard that can't survive the reload anyway.
- **`main.tsx`-only wiring** — intentional and matches the design doc's stated scope (admin/publisher pages out of scope). Confirmed deliberate.

Full frontend suite green (407 tests) at HEAD.

bbernstein:
## ✅ Ready for merge

- **Claude review** (substantive finding): `data/weekly-themes/*.json` served immutable → fixed in `c6d10a3` (excluded from immutable Pass 1, uploaded `no-cache` in both deploy paths). The follow-up Claude review re-ran and found **no new bugs**, confirming the gap is closed and the design doc is accurate.
- **PR body** upload-order wording corrected to state the real invariant.
- **Copilot** could not review (reported its own account quota limit); it is an auto-review app and cannot be re-requested manually.

**Threads to manually resolve:** none — there are no unresolved review threads (0 open).

**Non-blocking items intentionally deferred** (rationale in the [reply above](https://github.com/bbernstein/chq-calendar/pull/137#issuecomment-4992136436)): dropped `*.xml`/`*.txt` special-case (no such files exist today), `sessionStorage` silent-no-op edge case (flagged awareness-only, can't survive a reload anyway), and `main.tsx`-only scope (matches the design doc).

**Checks:** all passing — `test-backend (24/25)`, `test-frontend (24/25)`, `claude-review`. `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

The PR is ready for merge. Leaving the actual merge to @bbernstein.

PR #210, opened 2026-08-10
fix: repair local Docker dev setup, and guard the production deploy on forks

Woodwell:
Supersedes #208 (closed) — this branch contains that commit plus three more.

## What changed

Four independent fixes, all aimed at making a fresh clone actually runnable:

**1. Local Docker dev setup — workspace lockfile resolution** (`3410c07`)

Both `Dockerfile.dev` files built with a workspace-scoped context, isolated from the repo-root `package-lock.json` this npm-workspaces monorepo relies on:

- backend: `npm install` couldn't resolve `@chq-calendar/publisher-format` (a local workspace package, never published) and 404'd trying to fetch it from the registry.
- frontend: with no lockfile in context, npm re-resolved semver ranges fresh, landing on `@preact/preset-vite@2.10.5` instead of the pinned `2.10.3` — 2.10.5 pulls in the ESM-only `zimmerframe`, which breaks its hook-name transform plugin and blanks the page.

Both Dockerfiles now build from the repo root with the real lockfile, and `docker-compose.yml` mounts the whole repo with anonymous volumes protecting each container's own installed `node_modules`. Adds a `.dockerignore` so the root build context skips `node_modules`/`.git`/`docs`/`infrastructure`, and an `allowScripts` stanza in the root `package.json` covering the four postinstalls npm's allow-scripts policy flags (esbuild ×2, `@tailwindcss/oxide`, `aws-sdk`).

**2. Idempotent, runnable local setup script** (`d9337dc`)

- `backend/Dockerfile.dev`'s start command builds `tools/publisher-format` inside the container as root, directly into the bind-mounted host source tree, leaving `dist/` root-owned — a subsequent host-side `npm install` then failed with EPERM. Now chowns `dist/` back to whoever owns the bind-mounted source.
- `setup-local.sh`'s `check_port()` used `lsof -Pi ... -sTCP:LISTEN`, which can't see Docker's root-owned port-forwarding sockets when run as a non-root user, so real conflicts slipped through silently. Replaced with a `/dev/tcp` connect check that also treats already-running `chq-calendar-*` containers as safe to proceed past, so re-running against a live stack is a no-op instead of a hard exit.
- `package.json`'s `setup` script pointed at `scripts/setup.sh`, which doesn't exist — only `setup-local.sh` does. `npm run setup` failed immediately.

**3. Stale Node.js prereq in README** (`5a1ef5b`)

README said "Node.js 22+ (minimum 20.19)", but `.nvmrc` pins 24 and `engines.node` is `>=24.0.0` in every workspace. Not updated since the baseline moved to 24 (`7a2670c`).

**4. Skip the production deploy job on forks** (`70ea614`)

`deploy-production.yml` triggers on push to `main` and deploys straight to chqcal.org using static AWS keys, with no branch-protection gate on the `environment: production` target. On a fork it red-X's every push to `main` since no secrets are configured — noisy but harmless *today*. The real risk: if a fork owner ever added real deploy credentials to their fork's secrets, the next push to their `main` would deploy over the live production site. Adds `if: github.repository == 'bbernstein/chq-calendar'` so the job cleanly skips on forks.

## Why

Cloning the repo fresh and running the documented local dev setup did not work on either macOS or Linux.

## Verification

- [x] `docker compose up -d --build` — all four services healthy, backend `/health` OK, frontend renders the calendar UI in-browser
- [x] `npm run setup` run twice against an already-running stack — second run completed cleanly, no EPERM, no false port conflict, `tools/publisher-format/dist` stayed host-owned
- [x] `npm run build --workspace=frontend`
- [N/A] `npm run validate --workspace=backend` (no backend source changed)
- [N/A] `cd ios && xcodebuild test ...` (no iOS changes)
- [ ] Confirm the deploy job still runs normally on push to `bbernstein/chq-calendar:main`

## App Store listing

[skip-screenshots: no iOS or user-visible UI change — local dev tooling, README, and CI only]

- [N/A] Screenshots regenerated and manifest committed
- [N/A] `docs/app-store/listing-copy.md` re-read for invalidated claims
- [x] Not applicable — no visual change

## Review notes

Open feedback in [this review comment](https://github.com/bbernstein/chq-calendar/pull/210#issuecomment-5246561094) — most items are follow-ups rather than blockers, but items 1, 2 and 6 are worth resolving before merge.


.github/workflows/deploy-production.yml
```diff
@@ -14,6 +14,12 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     environment: production
+    # Guards against forks: this job deploys straight to chqcal.org on push
+    # to main with static AWS keys and no branch-protection gate. A fork
+    # would otherwise red-X on every push (no secrets configured) or, worse,
+    # actually deploy over production if someone ever added real creds to
+    # the fork's secrets.
+    if: github.repository == 'bbernstein/chq-calendar'
```
bbernstein, line 22: Good idea to add that check

Woodwell: thx.   Looks like I need to fix claude code review hooks at my end before CI will complete.

Woodwell: claude thinks this is a general setup issue, see issue 211

Issue #226, opened 2026-08-15
iOS: step forward/back by a day or a week from Today, This Week, and the Events list


bbernstein:
Companion to #225 (the web version). Deliberately filed separately so the two can be worked independently — the gaps are real on both platforms but the remaining work is very different, because iOS already has most of the data model.

Scope note: this is **not** a request for a date picker or "jump to next Tuesday." The ask is **relative stepping** — forward and back by one day or one week from wherever you already are.

## Problem

Three places on iOS where you can't move by a day or a week:

1. **Events tab, `.today` scope** — no way to reach tomorrow or yesterday. Same gap as web.
2. **Events tab, `.thisWeek` / week selection** — no way to reach next or previous week without reopening the date sheet and re-picking on the strip.
3. **My Day** — you can move a day by *tapping an adjacent chip*, but there is no ±1 stepper, and no week movement at all. The only fast-travel controls are the expand chevrons, which jump to the **season edge** in one tap (`MyDayView.swift:173-181`, `:199-207`) — a "go far" affordance with no "go one" complement.

`showNextDay()` has the same `.next`-only limitation as web:

```swift
// ios/ChqCalendar/App/AppModel.swift:1234-1236
func showNextDay() {
    filter.extraDays += 1
}
```

gated in the UI at `EventListView.swift:180-184`:

```swift
if model.isCurrentYear && model.filter.dateScope == .next {
    Button("Show next day") { model.showNextDay() }
}
```

## iOS is much closer to done than web

This is the key asymmetry and the reason for a separate issue. The web has **no state field representing the day being viewed**. iOS already does:

- **`DateScope.day`** — an absolute single-day scope carrying `selectedDayKey` (`ChqCalendarShared/Data/UserStateStore.swift:23`, `:56`), documented as "derived, not pickable."
- **`browseDay(_:)`** (`AppModel.swift:1074-1085`) already sets the scope, normalizes the key through `ChqTime.dayKey(for:)`, and clears `selectedWeeks` + `extraDays` in one assignment.
- **`ChqTime.day(_:offsetBy:)`** (`ChqCalendarShared/Support/ChqTime.swift:207-214`) already does DST-safe ±N day arithmetic via `Calendar`, not 86400-second math.

So day stepping on the Events tab is close to:

```swift
guard let key = filter.selectedDayKey ?? currentAnchorDayKey,
      let stepped = ChqTime.day(key, offsetBy: delta) else { return }
browseDay(stepped)
```

[...]

Issue #225, opened 2026-08-15
web: no way to move forward or back by a day or a week from Today / This Week


bbernstein:
## Problem

In **Now** mode there is a "Show next day" button. In **Today** mode there is nothing — no way to reach tomorrow. Same for **This Week**: no way to reach next week without going to the week strip and re-picking.

Scope note: this is deliberately *not* a request for a date picker or "jump to next Tuesday." The ask is **relative stepping** — forward and back by one day or one week from wherever you already are.

## Why it can't be done today

`'today'` is written as a predicate with no offset parameter:

```ts
// frontend/src/lib/utils/dateHelpers.ts:52-56
eventDate.toDateString() === today.toDateString()
```

and `extraDays` — the state behind "Show next day" — is hard-gated to `'next'` in three places:

- `frontend/src/app/page.tsx:81` — `adaptiveEndDate` returns `undefined` for any other mode, and it's the only thing `extraDays` feeds
- `frontend/src/components/calendar/EventList.tsx:118` — the button is gated on `dateFilter === 'next'`
- `frontend/src/hooks/useFilterState.ts:209-213` — reset on every `dateFilter` change

There is no state field representing "the day being viewed." `FilterState` (`useFilterState.ts:6-19`) has `dateFilter`, `selectedWeeks`, and `extraDays`, nothing else date-bearing. The mode union is `'all' | 'today' | 'next' | 'this-week'` (`useFilterState.ts:4`).

## Widening vs. moving — these are different operations

Worth separating before designing anything. "Show next day" **widens**: `extraDays: state.extraDays + 1` (`useFilterState.ts:85`) pushes the *end* boundary out while the start stays put, so the list accumulates. Stepping **moves**: both boundaries shift together and the previous day leaves the list.

Only widening exists today. Both are plausibly wanted — "show me one more day of what's coming" is not the same request as "show me tomorrow instead." Any design here should say which one each control does, and possibly offer both.

## What iOS already has

iOS is meaningfully ahead here and its design questions are already litigated:

- **`DateScope.day`** — an absolute single-day filter carrying `selectedDayKey` (`ios/ChqCalendarShared/Data/UserStateStore.swift:23`, `:56`). Documented as "derived, not pickable"; it arrives only from My Day's empty-day "Browse …" action. The web has no equivalent.
[...]

Issue #156, opened 2026-08-02
iOS: "Show next day" (extraDays) not reset when date scope changes


bbernstein:
## Problem

`FilterSelection.extraDays` (incremented by the "Show next day" button
shown when `.next` scope has run out of events, via
`AppModel.showNextDay()`) is not reset when the date scope changes via
`selectScope`. Tapping "Show next day" a few times to widen the visible
window, then changing scope (e.g. Now → This Week → Now again) leaves
`extraDays` at whatever it was, so the window is wider than a fresh
`.next` selection would be — with no visible indication to the user of
why they're seeing more days than expected.

## Expected behavior

`extraDays` should reset to 0 whenever the date scope changes via
`selectScope` (and probably also on `clearAll`/`clearNonDateFilters`,
though it's already implicitly 0 after those since they rebuild the
selection from scratch — worth confirming). "Show next day" widening the
window should only persist within a single, continuous `.next`-scope
session.

## Context

- Source: `docs/plans/2026-08-02-ios-ux-filtering-follow-ups.md`, §4
  ("pre-existing" — predates PR #151, not introduced by it, but
  surfaced during that branch's work).
- Relevant code: `ios/ChqCalendar/App/AppModel.swift`
  (`selectScope`, `showNextDay`, `FilterSelection.extraDays`).

🤖 Filed with [Claude Code](https://claude.com/claude-code)

PR #233, opened 2026-08-16
refactor(ios): derive one date window and one owner for the scope downgrade

bbernstein:
Phase 1b of the cross-platform date-navigation initiative (#225 web / #226 iOS). The web half merged yesterday as #232; this is the iOS mirror. Phase 0 (the area-split deploy) was #231.

Plan: `docs/superpowers/plans/2026-08-15-date-navigation-phase-1b-ios-window-model.md`
Spec: `docs/superpowers/specs/2026-08-15-cross-platform-date-navigation-design.md`

[skip-screenshots: phase 1b is a pure refactor; no visible change — button label, placement and behavior are unchanged]

## What changes

**`EffectiveScope` is now the single owner of two rules** that were restated in four places: a past or future season has no "now", so relative scopes degrade to `.all`; and `.day` is exempt because it names an absolute calendar date — but only while it carries a key, since a keyless `.day` filters nothing, which is exactly what `.all` means. `EventFilter`, `DateFilterLabel`, `FilterChipState` and `EventListView` each restated some part of that by hand, with no compiler help if they drifted. `FilterChipState` is the one where drift is a **silent** wrong answer: a chip renders selected against a filter doing something else.

**`ViewWindow` replaces `EventFilter`'s six-branch date `switch` with one range check.** It mirrors the web's `dayWindow.ts`: half-open (`start <= x < endExclusive`, carried as a `Range<Date>` so the type enforces it), day keys derived from the range rather than alongside it, expansion that only ever grows, and a `bounds` clamp applied to the expansion *inputs*.

**`extraDays: Int` is deleted** in favour of `windowStartDayKey`/`windowEndDayKey`. It was only ever the window's end expressed as an offset meaningful under one scope, which is why `clearScopeLocalDateState` existed to sweep it up on every scope change. Absolute day keys make the state impossible to misapply rather than merely tidied — the #156 bug class closed by construction. Both fields are session-only; `saveFilters` builds a separate `PersistedFilters` that omits them, so the on-disk schema is untouched.

No rail, no stepping, no scope-set change — those are phases 2–4. iOS's scope set is already the target one, so unlike the web it has nothing to change there.

## Behavior

Zero user-visible change. The "Show next day" button keeps its label, placement and visibility condition in every reachable state; only its call site is renamed and its condition rewritten to an equivalent expression (proven equivalent case-by-case, including `.day`, keyless `.day`, and archived years).

One deliberate delta: `expandWindowEnd` is bounded by the navigable range where `extraDays += 1` was unbounded. This can never hide an event — `navigableBounds.upperBound` is by definition the max over every event's day key, so any day it withholds provably has no event. The button becomes inert exactly where it previously did nothing visible.

## Three defects in the plan, caught before merge

The plan is a session artifact, and this branch corrects it where it was wrong:

1. **Its own tests referenced a `w.end` property its `ViewWindow` never defines**, and two asserted an inclusive-minus-epsilon relationship its own Global Constraints forbid. Those tests now assert exact equality on `endExclusive` — strictly stronger, since `.season` and `.thisWeek` carry `SeasonWeek`'s bounds through verbatim.
2. **Its `EventFilter` line was `$0.start >= window.start && $0.start <= window.end`** — the phantom property again, plus an inclusive comparison on an exclusive bound. Now `window.contains($0.start)`, so the rule lives in one place.
3. **Its `FilterChipState.isScopeSelected` dropped an `isCurrentYear` gate** on the "only the current week is selected" equivalence. A pre-existing test caught it. Restored, and the collapse was then verified cell-by-cell against the deleted block across every combination of scope, selection, week and year axis.

`WindowParityTests` also shipped from the plan containing a test that **could not fail** — it asserted that `Array<String>.sorted()` sorts strings, in the file meant to prove this phase's whole point. It now round-trips through `ChqTime.dayKey` and exercises `ViewWindow` directly, citing the web's mirror tests by name. The plan is corrected so the next reader does not regenerate it.

## Cross-platform parity

The two implementations agree on day-key format and ordering, half-openness, `lastDayCovered`'s midnight rule, growth-only expansion, input-clamping, and DST-safe day arithmetic. `WindowParityTests.swift` pins those, naming the web's mirror in `frontend/src/__tests__/lib/utils/dayWindow.test.ts`.

One real divergence is documented rather than papered over: **`this-week` out of season returns `null` on the web** (list shows nothing) **but a rolling 7-day window on iOS**. Both preserve their own platform's prior behavior, so neither changes here; phase 3 reconciles it.

## Mutation proofs

Every load-bearing invariant was verified by breaking it and watching the right tests fail:

- Deleting the `.day` exemption → 3 characterization tests fail.
- `dayAfter` returning `ChqTime.endOfDay` (`23:59:59.000`), and returning `next - 0.001` → 6 tests fail, including the day-tiling one. The second is the mutation that matters: it is exact on the web, where `Date` is integer milliseconds, and wrong here, where `Date` wraps a `Double`.
- Switching the clamp to the rejected clamp-after-merge form → a real `Fatal error: Range requires lowerBound <= upperBound`. Worth noting the obvious test did **not** catch this: only a base sitting *outside* the bounds discriminates the two, which is the off-season case.

## Tests

789 tests, 57 suites. The 14-test characterization matrix written before any refactor passed unmodified at every commit, and no pre-existing test was deleted — the ten that named `extraDays`/`showNextDay` were translated, two of them to stronger assertions.

**One known-red test, pre-existing and environmental:** `AppGroupTests.containerURLIsNilInTheUnitTestHost` asserts that no App Group container exists in the unit-test host. On a machine whose simulator has the container provisioned, that premise is false. Verified by running it against `main` at `5d3e169` in a clean worktree, where it fails identically. It should pass on CI, whose runners are freshly provisioned — worth a follow-up to make the test assert something environment-independent.

## Deferred to phase 2

`ViewWindow.make`'s `bounds` parameter defines both the expansion clamp and the `.all` scope's day-key projection, and `EventFilter` passes cheaper season-only bounds when no expansion is active (skipping an O(n) `DateFormatter` scan the common path discards). That is invisible today because `EventFilter` reads only `contains`, and it is now pinned by a contract test — but phase 2's day rail is the first caller that will read `startDay...endDay`, and should revisit the seam.

https://claude.ai/code/session_014PN8JR7jFsQuzmhnrLxDDT

### DIFF
--- a/docs/superpowers/plans/2026-08-15-date-navigation-phase-1b-ios-window-model.md
+++ b/docs/superpowers/plans/2026-08-15-date-navigation-phase-1b-ios-window-model.md
@@ -23,7 +23,7 @@
 - **`windowStartDayKey` / `windowEndDayKey` are session-only.** `UserStateStore.saveFilters` builds a separate `PersistedFilters` struct (`UserStateStore.swift:279-290`) that already omits `searchText`, `extraDays`, and `selectedDayKey` — so **no persistence change is required**, and none may be added. They must also be excluded from `FilterSelection.isDefault`.
 - **The `.day` exemption is the highest-risk change in this plan.** It currently lives in three places that must agree: `EventFilter.swift:42`, `DateFilterLabel.swift:74-78`, and `FilterChipState.swift:56, 66, 92, 101`. `FilterChipState` is the one where getting it wrong is a **silent behavioral bug** rather than a compile error. Task 1 pins all three before Task 2 touches any of them.
 - **Screenshot obligation.** This plan modifies `ios/ChqCalendar/Features/Calendar/EventListView.swift` and `ios/ChqCalendarShared/**`, both matched by `.github/workflows/app-store-assets.yml`. **No pixel changes** — the "Show next day" button keeps its label, placement, and behavior. Opt out explicitly in the PR description with `[skip-screenshots: phase 1b is a pure refactor; no visible change — button label, placement and behavior are unchanged]`.
-- **Version is 1.1.3** wherever a version is referenced. Do not reintroduce 1.1.2.
+- **This plan does not touch `MARKETING_VERSION`.** `project.pbxproj` is at `1.1.2` on `main` as of this writing; that is current, not stale. If a version bump is ever needed alongside this work, confirm the live value in `project.pbxproj` first rather than trusting a number written into this plan on an earlier day.
 
 ---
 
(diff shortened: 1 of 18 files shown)


bbernstein:
All review comments have been addressed — no code changes were required.

**Threads to mark resolved:** none. Neither reviewer left an inline comment, so there are no conversations open.

**Review status:**
- **Copilot** reviewed 18 of 18 files and generated no comments. I checked for a collapsed `Suppressed comments` block — there isn't one.
- **`@claude`**: *"No blocking issues found."* It independently confirmed the four things this PR most needed confirmed: that `EffectiveScope.resolve` is idempotent and matches what all three consumers need; that clamping the expansion *inputs* is correct because `base` can legitimately sit outside `bounds`, and that the alternative would produce an inverted `Range` and crash; that the day-key string comparisons are safe because `ChqTime.dayKey` is fixed-width; and that `EventListView`'s collapsed button condition is genuinely equivalent to the old one.

**Its two non-blocking observations, and why neither changes this PR:**

1. **`windowStartDayKey` has no production writer yet.** Correct and deliberate — there is no `expandWindowStart()` until the day rail lands in phase 2. The field is added now so `ViewWindow` and `FilterSelection` reach their final shape in one behavior-preserving change rather than two.
2. **`AppModel.selectScope(_:)`'s early-return skips `clearScopeLocalDateState()` when you re-tap the already-active scope**, so a `.next` window that has been expanded keeps its expansion. The reviewer correctly identifies this as pre-existing — `selectScope` is untouched by this branch and `extraDays` behaved identically. Fixing it here would be a behavior change in a PR whose entire premise is that behavior is preserved, so it belongs in phase 2 alongside the rail and `⟳ Now`, where the expansion state actually becomes user-visible. **Worth filing as a follow-up** — say the word and I will.

**Checks:** all passing — `Build and test (iOS)`, `test-frontend (24)`, `test-frontend (25)`, `test-backend (24)`, `test-backend (25)`, `claude-review`, `screenshot-freshness`. `mergeStateStatus` is `CLEAN`.

**On the one locally-red test:** `Build and test (iOS)` passing on CI settles it. `AppGroupTests.containerURLIsNilInTheUnitTestHost` fails on my machine because this simulator has the App Group container provisioned, making the test's premise false locally — I verified it fails identically on `main` at `5d3e169` in a clean worktree, and it passes here on a freshly-provisioned runner. That test asserts an environment property rather than a code property; making it environment-independent is a reasonable follow-up.

This PR is ready for merge. Leaving the merge itself to you.

bbernstein:
Filed the two follow-ups from this review:

- #234 — re-tapping the active scope keeps a widened date window (`AppModel.selectScope`'s early return skips `clearScopeLocalDateState()`). Pre-existing; same bug class as #156, and worth deciding as part of phase 2's design, since the rail and `⟳ Now` make "reset the scope I'm on" a real user intent.
- #235 — `AppGroupTests.containerURLIsNilInTheUnitTestHost` asserts a property of the machine rather than of the code, with the `main`-worktree evidence and three candidate fixes.

PR #236, opened 2026-08-16
feat(web): grow the list past the edges of the scope, behind VITE_NAV_V2

bbernstein:
## What changed

The event list can grow past the edges of the current date scope: forward automatically as you scroll, backward through an explicit "Show earlier" control that preserves your scroll position. All of it is behind `VITE_NAV_V2`, which is **unset everywhere**, so this merges dark and production behaviour is unchanged.

## Why

Phase 2 of the cross-platform date-navigation initiative ([spec](../blob/feat/date-nav-phase-2-web-render-window/docs/superpowers/specs/2026-08-15-cross-platform-date-navigation-design.md), phases 0/1a/1b already merged as #231/#232/#233). The underlying complaint is that *date is modelled as a filter, and filters are walls* — reaching the edge of what you asked for is a dead end escapable only by asking a different question. Phase 1 replaced four scope-specific date predicates with one derived `ViewWindow`; this phase makes that window movable.

Phase 3 owns the day rail, the scope-set change (drop "This Week", add "All Season"), the active-filter date chip naming the real window, and the About-guide copy. The flag exists precisely because "Today" can now show two days while the chip still says "Today".

### The design, in one paragraph

There are **two** windows and they must not be conflated. The **view window** is data — which days pass the date filter, owned by `useFilterState`, derived by `dayWindow.ts`; growing it costs a refilter of ~1,470 events. The **render window** is DOM — which of those days are mounted; growing it costs a render. The bottom sentinel does both jobs in strict order: extend the render window while loaded days remain, and only when none do, ask the page to widen the view window to **the next day that has events under the current non-date filters** (not the next calendar day — that would move the edge and add nothing whenever the next day is empty or filtered out). The render window is anchored on a **day key, never an index**, because expanding backward prepends day groups and shifts every index underneath it.

## Verification

- [x] `npm run build --workspace=frontend` — green. **769 tests / 71 files**, line coverage **79.26%** against the 74.3 floor.
- [x] `npm run validate --workspace=backend` — green (backend untouched; run per the checklist).
- [x] `npm run type-check`, `npm run lint` — clean, zero warnings on all changed source files.

### Browser pass, flag ON (`VITE_NAV_V2=true npm run dev`, Playwright-driven)

Two of this phase's three biggest risks are invisible to jsdom by construction, so these are the only place they are checked.

| # | Check | Result |
|---|---|---|
| 1 | Default (`Now`) scope loads and reads as before | pass |
| 2 | Scroll to bottom keeps loading days, no button; old "Show next day" gone | pass |
| 3 | Loading stops at the season end, no spin | pass |
| 4 | `Today` extends into following days on scroll | pass |
| 5 | A short list does **not** expand on its own | pass — see note below |
| 6 | **"Show earlier" does not move what you are looking at** | pass — measured |
| 7 | Repeated "Show earlier" to the season start; button disappears | pass — 72 clicks, zero dropped days |
| 8 | Changing a filter resets to the top | pass |
| 9 | Starring with favourites-only OFF does not jump the list | pass |
| 10 | Un-starring with favourites-only ON resets | pass (intended trade) |
| 11 | Year switch drops time-relative scopes and expansion state | partial — filter reconciliation confirmed; 2025 content unverifiable, local dev has no 2025 data by design |
| 12 | Mobile 375px: sticky headers correct, control reachable | pass |

Flag OFF: "Show next day" is back, the list grows 50 events at a time, nothing named "Show earlier" exists.

### Two bugs the browser found that no test could

**Check 6 failed the first time, with a real defect.** The render anchor was only latched by the first *downward* growth step, so pressing "Show earlier" before ever scrolling down — an ordinary first action — re-ran the initial fill over the newly prepended array and silently unmounted days already on screen. Measured: a **3,903px position jump with an entire 24-event day dropped**. After the fix, the same case is a **58.5px drift with the day preserved**. A regression test now reproduces it in jsdom (it fails against the old code; the existing prepend test could not catch it, because it triggers growth first and that sets the anchor as a side effect).

**The short-list guard was then found to be a proxy for the wrong thing.** `scrollHeight > innerHeight` is satisfied by page chrome alone; the guard was never demonstrated by any evidence on the branch (check 5 originally passed because `canExpandEnd` was false, not because of the guard). It now reads live scroll position, re-verified in the browser against a configuration that genuinely has later-day events. Closing that opened a further hole — an `IntersectionObserver` reports intersection *changes*, so a sentinel already intersecting at mount never reports again — fixed by re-arming the observer on the reader's first scroll, and a subsequent session-wide-latch regression was caught in re-review and **reproduced live by stashing the fix back out on a running dev server**.

## Known residuals, deliberately deferred

- **~55–59px scroll drift per "Show earlier" click.** Two orders of magnitude smaller than the 3,903px defect it replaced. Root cause is *not* the button relabelling (a label change moves no box) — the correction compares **total document height** before and after, which assumes every height change is above the reader and has landed by the time the layout effect runs; the instrumentation shows the document still growing ~300ms later. The durable fix is rect-based (capture a stable node's `getBoundingClientRect().top`, then `scrollBy` the delta), which stays correct when phase 3 inserts a sticky day rail above the day headers — and which the current approach will handle *worse*, not better. Recorded in the spec for phase 3.
- **Expansion granularity is one event-day per refilter.** Converges; under a sparse filter it takes ~4–6 steps to push the sentinel clear. A target of "next day adding at least a batch" would cut it to roughly one.
- **An empty scope is still a dead end** — zero matches renders `EmptyState`, so there is no sentinel and no control. Phase 3's rail is scope-independent and fixes this.
- The layout-effect **timing** of the scroll correction is verified only in the browser; jsdom does not paint and cannot distinguish `useLayoutEffect` from `useEffect`.
- The flag-dispatch tests live in `EventList.legacy.test.tsx` — they must survive that file's deletion in phase 3.

## Notes for review

- `EventList.tsx` is now a dispatcher; the day-group JSX moved **verbatim** into `EventListView` and the old container into `EventListLegacy`. Flag-off equivalence was verified structurally — diffing the pre-branch `EventList.tsx` against `EventListLegacy.tsx` shows a pure extraction — and is pinned by 7 characterization tests written against the unmodified component *before* it was touched, which have passed unedited through every commit since.
- `navEventDays` deliberately depends on `nonDateFilterOpts`, **not** `filterOpts`, so the second full filter pass runs on filter changes and not once per window expansion.
- `eventHelpers.localDateKey` was deleted in favour of `dayWindow.dayKeyOf`. The duplication was harmless before this branch and load-bearing after it: the anchor matches `DayGroup.key` while navigation targets compare `dayKeyOf` output, so drift would have stopped the list growing with nothing failing.

## App Store listing

- [x] Not applicable — web-only change, no `ios/` path touched.

https://claude.ai/code/session_011zuRAidtZqiiaa7vhFpFwb

### DIFF
--- a/docs/superpowers/plans/2026-08-15-date-navigation-phase-2-web-render-window.md
+++ b/docs/superpowers/plans/2026-08-15-date-navigation-phase-2-web-render-window.md
@@ -2182,7 +2182,7 @@ construction.
   (status line only)
 - Modify: this plan (check off completed tasks)
 
-- [ ] **Step 1: Full verification sweep**
+- [x] **Step 1: Full verification sweep**
 
 From the repo root:
 
@@ -2195,7 +2195,16 @@ Expected: both green; frontend line coverage at or above the 74.3 floor in
 `.coverage-floor.json`. If coverage dropped, add tests — do not lower the
 floor.
 
-- [ ] **Step 2: Browser pass with the flag ON**
+- [x] **Step 2: Browser pass with the flag ON** — performed; 11 of 12 checks
+  passed, check 6 found a real bug (see task-7-report.md): the render-window
+  `anchor` in `EventListWindowed` starts `null` and is only ever set by the
+  bottom-sentinel's growth callback. If the reader clicks "Show earlier"
+  before any bottom-scroll growth has fired, `renderEndIndex` recomputes
+  `fillFrom(groups, 0, 50)` from scratch against the newly-prepended array and
+  can silently drop previously-rendered later days from the DOM (confirmed
+  reproducible, confirmed it does NOT occur once a growth step has fired
+  first, confirmed it compounds across repeated clicks). Must be fixed before
+  `VITE_NAV_V2` is flipped on.
 
 ```bash
 cd frontend && VITE_NAV_V2=true npm run dev
@@ -2231,7 +2240,9 @@ Visit `http://localhost:3000` and confirm each of these, in order:
 Record the result of every numbered check in the PR body. If the Chrome
 extension is unavailable, do the pass manually and say so.
 
-- [ ] **Step 3: Browser pass with the flag OFF**
+- [x] **Step 3: Browser pass with the flag OFF** — confirmed byte-identical
+  legacy behaviour: "Show next day" reappears, batch growth is exactly 50 at
+  a time, no "Show earlier" anywhere.
 
 ```bash
 cd frontend && npm run dev
@@ -2240,14 +2251,15 @@ cd frontend && npm run dev
 Confirm: "Show next day" is back under the `Now` scope, the list grows 50
 events at a time, and nothing named "Show earlier" exists anywhere.
 
-- [ ] **Step 4: Update the spec's status line**
+- [x] **Step 4: Update the spec's status line**
 
 In `docs/superpowers/specs/2026-08-15-cross-platform-date-navigation-design.md`,
 change the `**Status:**` line to record that phases 0, 1a, 1b are merged and
 phase 2 is implemented behind `VITE_NAV_V2` (default off), and that the flip
 is phase 3's first act.
 
-- [ ] **Step 5: Commit and push**
+- [x] **Step 5: Commit** — committed locally only, per the controlling
+  agent's instruction to stop before Step 6 (no push, no PR).
 
 ```bash
 git add docs/superpowers/specs/2026-08-15-cross-platform-date-navigation-design.md docs/superpowers/plans/2026-08-15-date-navigation-phase-2-web-render-window.md
--- a/docs/superpowers/plans/2026-08-15-date-navigation-phase-2-web-render-window.md
+++ b/docs/superpowers/plans/2026-08-15-date-navigation-phase-2-web-render-window.md
@@ -1151,8 +1151,13 @@ export interface EventListProps extends Omit<EventListViewProps, 'groups'> {
 
 // The windowed container arrives in the next task; until then the flag
 // selects nothing and the legacy path is unconditional. The unused names are
-// destructured so they do not reach `rest` — `@typescript-eslint/no-unused-vars`
-// ignores rest siblings, so this lints clean without a suppression.
+// destructured so they do not reach `rest` and get spread onto a component
+// whose props do not declare them. `@typescript-eslint/no-unused-vars` counts
+// a destructured parameter as an argument, so this reports six warnings until
+// Task 4 gives all six a use — accepted rather than suppressed, because a
+// suppression would outlive the reason for it. Frontend lint does not fail on
+// warnings; Task 4 must end with `npx eslint src/components/calendar/EventList.tsx`
+// clean.
 export function EventList({ navV2, resetKey, earlierDay, onShowEarlier, canExpandEnd, onExpandEnd, ...rest }: EventListProps) {
   return <EventListLegacy {...rest} />;
 }
--- a/docs/superpowers/plans/2026-08-15-date-navigation-phase-2-web-render-window.md
+++ b/docs/superpowers/plans/2026-08-15-date-navigation-phase-2-web-render-window.md
@@ -2262,8 +2262,8 @@ floor.
   mount). Also fixed: a blocked sentinel no longer shows a permanent
   "Loading more events…"; `eventHelpers.ts`'s day-key format is no longer a
   second implementation of `dayWindow.ts`'s; the spec's status line no
-  longer claims "merged" (no PR yet) or a clean browser pass (one residual
-  is recorded); the residual's root cause is corrected and the durable fix
+  longer claims "merged" or a clean browser pass (one residual is
+  recorded); the residual's root cause is corrected and the durable fix
   recorded for phase 3.
 
 ```bash
--- a/docs/superpowers/plans/2026-08-15-date-navigation-phase-2-web-render-window.md
+++ b/docs/superpowers/plans/2026-08-15-date-navigation-phase-2-web-render-window.md
@@ -1354,6 +1354,44 @@ describe('EventListWindowed', () => {
     expect(onExpandEnd).not.toHaveBeenCalled();
   });
 
+  it('still mounts more loaded days on a page that cannot scroll', () => {
+    // The scrollable-page guard belongs to the expensive step only. Without
+    // this case, an implementation that wrapped BOTH steps in the guard would
+    // pass every other test in this file, because they all force a scrollable
+    // page — and the list would refuse to grow into days it had already
+    // loaded whenever the viewport was taller than the content.
+    setPageScrollable(false);
+    const groups = [group('2026-07-05', 60), group('2026-07-06', 20)];
+    render(<EventListWindowed {...baseProps} groupedEvents={groups} />);
+    expect(screen.queryByText('Day 2026-07-06')).not.toBeInTheDocument();
+
+    io.trigger();
+
+    expect(screen.getByText('Day 2026-07-06')).toBeInTheDocument();
+  });
+
+  it('re-anchors on a filter change even when the day groups are rebuilt', () => {
+    // The existing reset test reuses the same array reference, so it cannot
+    // tell a derived anchor from an effect-synchronised one. Neither can this
+    // one, in the end: `rerender` runs inside `act`, which flushes effects
+    // before returning, so the extra frame an effect would paint is invisible
+    // here. What this DOES pin is that re-anchoring survives the realistic
+    // shape of a filter change — a brand-new array whose surviving day keys
+    // overlap the old anchor — which is the state that makes the effect
+    // version misbehave in a browser.
+    //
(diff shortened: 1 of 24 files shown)


bbernstein:
@claude Please re-review. Since your last pass I've addressed all four of your observations plus Copilot's two:

1. **`renderResetKey` delimiter collision** — array fields now go through `JSON.stringify` instead of a `,` join, so `['a,b']` and `['a', 'b']` no longer produce the same key. Test added for exactly that pair.
2. **`SET_SELECTED_WEEKS` not clearing the window fields** — you were right to ask, and the honest answer is that it wasn't a deliberate decision. It now clears `windowStartDay`/`windowEndDay`, guarded on an actual content change for the same reason `SET_DATE_FILTER` is (`useScrollState` dispatches same-value week selections from several handlers). The guard has a regression test that was confirmed to fail against an unguarded implementation, not just asserted to.
3. **Observer churn on identity-only `groupedEvents` changes** — no change, deliberately. Recreating rather than getting cleverer is the design, and the component's doc comment records the caller-stability contract it rests on.
4. **A11y label divergence** — good catch, and it was a WCAG 2.5.3 "Label in Name" violation rather than a cosmetic difference: the accessible name ("Show earlier events, Friday, Jul 3") did not contain the visible text ("Show earlier (Friday, Jul 3)"). The `aria-label` is gone; the visible text is now the accessible name, which also removes the divergence from the legacy button you noted.

Copilot's two comments were both about doc sentences asserting PR state, which went stale the moment this PR opened — removed from the spec and the plan.

772 tests, 79.28% coverage against the 74.3 floor, legacy characterization tests still passing unedited.

bbernstein:
@claude Round 2 — Copilot's second pass reported "no new comments" in its summary while holding two real findings in a collapsed *Suppressed comments* block. Both are addressed in `df345b8`:

1. **`resetKey` was optional even when `navV2` was on**, with a `resetKey ?? ''` fallback in the dispatcher. Enabling the windowed path without it would have frozen the render window on a constant key — no reset across filter changes, wrong scroll position, and nothing failing anywhere. `EventListProps` is now a discriminated union on `navV2`: the `true` arm requires `resetKey`, and the fallback is gone. No casts were needed. `page.tsx`'s call site is split into two literal-`navV2` branches, since the runtime boolean it holds can't select a union arm. Pinned with a `@ts-expect-error` test, which fails type-check if the constraint is ever loosened — there is nothing to assert at runtime, because a violation never reaches `render()`.

2. **The unlabelled sentinel had zero height.** It now has `h-px`. Worth being precise about the status of this one: the empty-sentinel path *was* exercised in Chrome during verification and fired correctly (the short-list case went 1 → 6 events on the first scroll), and the IntersectionObserver spec does define behaviour for zero-area targets — but Safari and Firefox were never checked here, and a real pixel costs nothing. The labelled sentinel is unchanged.

The 7 characterization tests in `EventList.legacy.test.tsx` are still untouched — `git diff` on that file is empty, and both dispatch arms it exercises compile without edits, which is what I wanted the union to preserve.

774 tests, 79.3% coverage against the 74.3 floor.

bbernstein:
@claude Round 3. Copilot's third pass again reported "no new comments" while holding two findings in a collapsed *Suppressed comments* block — worth noting as a pattern, since the summary line is actively misleading. Both were real and are fixed in `2637df7`:

1. **`renderResetKey` was ambiguous.** `searchTerm` is free-form user input joined with `|` into a key whose other fields are JSON. The whole tuple is now `JSON.stringify`d rather than joined, which is unambiguous by construction rather than by luck. (Copilot's specific collision example doesn't literally collide under the current 8-field layout — the implementer checked numerically rather than taking my restatement of it on faith, and kept it as a test anyway.)

2. **The key disagreed with the search it describes**, so `"Organ"` and `"organ "` reset the render window despite `searchEvents` returning identical results. Rather than writing a second normalizer that could drift from the first, `searchHelpers` now exports `searchTermsOf` and both call it — the same single-source-of-truth move that `dayKeyOf` got earlier in this PR.

That second one had a trap worth recording, because a naive `trim()`-and-collapse fix would have introduced a bug: `searchEvents` splits on a **literal space**, not general whitespace, so `"a\tb"` is one term and `"a b"` is two — different results, and they must not share a key. And `""` early-returns *every* event while `"   "` is truthy, produces zero terms and matches *nothing* — opposites that both normalize to an empty array. All three behaviours are now pinned by tests.

Also closed the mirrored prop leak in the legacy dispatch arm that the windowed arm's fix left behind.

778 tests, 79.32% coverage against the 74.3 floor. `EventList.legacy.test.tsx` remains untouched.

bbernstein:
@claude Round 4 — `c7855fd`, comments only, no behaviour change.

Copilot's fourth pass (again "no new comments" with two findings in the suppressed block) caught a genuinely subtle documentation error, and it was mine. Two comments said `searchTermsOf` captures what `searchEvents` matches on, and justified `"a\tb"` ≠ `"a b"` with "a tab is not a split point". But `searchEvents` re-splits each term on `/\s+/` for its word-level scoring, so a tab **is** a separator there.

The conclusion survived — those two searches really are different — but the reason was wrong, and the belief a reader would take away (that tokenization describes matching) is false. That's worse than no comment in code someone will lean on.

Both now say the thing that is true *and* sufficient: tokenization determines **identity**, because every later stage of `searchEvents` is a pure function of the term list, and identity is the only property a reset key needs. The `searchTermsOf` doc now says outright that it is not a summary of matching semantics.

Worth noting for whoever reads this thread later: all four of Copilot's passes reported "generated no new comments" in the summary line while carrying real findings in a collapsed *Suppressed comments* block. Four for four. The summary is not a reliable signal on its own.

bbernstein:
## All review comments have been addressed

Five review passes (Copilot ×5, Claude ×4), four rounds of fixes. Every finding was real; none were disputed.

**Round 1 — `b6766ab`**
- `renderResetKey` collided across distinct filter states (`['a,b']` vs `['a','b']`).
- `SET_SELECTED_WEEKS` left the window expansion in place while every other filter reset it. Not a deliberate decision on my part — weeks feed the navigation targets, so it now clears, guarded against same-value re-dispatch the way `SET_DATE_FILTER` is. The guard was confirmed to fail against an unguarded implementation rather than assumed to matter.
- The "Show earlier" `aria-label` was a WCAG 2.5.3 violation, not a nicety: the accessible name didn't contain the visible text.
- Two doc sentences asserting PR state, removed.

**Round 2 — `df345b8`**
- `resetKey` was optional even with `navV2` on, with a `?? ''` fallback that would have frozen the render window on a constant key and failed nowhere. `EventListProps` is now a discriminated union; the `true` arm requires it. No casts. Pinned by a `@ts-expect-error` test that fails type-check if the constraint loosens.
- The unlabelled sentinel had zero height. Now `h-px` — it did work in Chrome during verification, but Safari and Firefox were never checked and a real pixel costs nothing.

**Round 3 — `2637df7`**
- `renderResetKey` is now one `JSON.stringify` of the tuple rather than a `|` join over free-form user input — unambiguous by construction rather than by luck.
- The key disagreed with the search it describes. Rather than write a second normalizer that could drift, `searchHelpers` now exports `searchTermsOf` and both call it. Three behaviours of `searchEvents` had to be preserved exactly and are now pinned by tests: it splits on a **literal space** (so `"a\tb"` is one term and `"a b"` is two), and `""` returns *every* event while `"   "` returns *none* — opposites that both tokenize to `[]`. A `trim()`-and-collapse fix would have introduced a bug.
- Closed the mirrored prop leak in the legacy dispatch arm.

**Round 4 — `c7855fd`** (comments only)
- Two comments overstated what `searchTermsOf` models, justifying a correct conclusion with a wrong reason. `searchEvents` re-splits each term on `/\s+/` for word-level scoring, so a tab *is* a separator there. Both now claim only what is true and sufficient: tokenization determines identity, which is all a reset key needs.

### Threads

Both inline threads are resolved and outdated. **Nothing needs manual resolution.**

### Checks

All passing on `c7855fd`: `test-frontend (24)`, `test-frontend (25)`, `test-backend (24)`, `test-backend (25)`, `claude-review`, `screenshot-freshness`. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, no pending reviewers, not behind `main`.

778 tests / 72 files, 79.32% line coverage against the 74.3 floor. The 7 characterization tests in `EventList.legacy.test.tsx` are byte-identical to when they were written against the pre-refactor component — which is the evidence that a flag-off build renders what `main` renders.

### Ready for merge

Merging is deliberately dark: `VITE_NAV_V2` is unset in CI and in the deploy workflow, so production behaviour does not change. Flipping the flag is a separate one-line follow-up, and phase 3 (day rail, scope-set change, the date chip naming the real window) is what makes the flag-on experience coherent — the chip still says "Today" while the window can show more than today, which is exactly why this ships off.

Deferred residuals are listed in the PR description; the ~55–59px scroll drift is the one worth carrying into phase 3, along with its corrected root cause and the rect-based fix that replaces it.

**Not merging automatically — that's yours to call.**

PR #239, opened 2026-08-17
feat(web): the filter panel leaves when you scroll

bbernstein:
Follow-up to #238. Phone testing found an **affordance mismatch** in the filter panel that PR shipped: it looks exactly like the page header, so readers expect scrolling to move it out of the way — and it didn't.

Measured before this branch: after tapping the last chip of the season you land ~31,000px down, and the panel that gets you back to the filters then refuses to behave like the header it resembles.

## What changed

**A scroll gesture dismisses it**, animated, and only the Filters control brings it back. Scrolling back up shows results, not the panel.

**Dismissal keys off input events — `wheel`, `touchstart`, `touchmove`, `mousedown`, `keydown` — and never the `scroll` event.** That is load-bearing, not stylistic: the panel's own opening correction calls `scrollBy`, and a filter change reflows the list. A `scroll` listener would dismiss the panel in the frame it opened, and again on the reader's first venue tick.

**It animates out of flow.** On dismiss the panel leaves flow immediately so layout settles once and one scroll correction runs; the same element then animates at `position: fixed` using only `transform` and `opacity`, unmounted on transition end. Animating height would relayout hundreds of event cards every frame and leave the correction chasing a moving target.

**A caret** closes it without scrolling, **a funnel icon** replaces the word on a crowded rail, and **a dot** marks active filters — lit when you've changed something from how the page starts, not on a default visit.

Design and the rejected alternatives: `docs/superpowers/specs/2026-08-17-web-filter-panel-dismissal-design.md`.

## The defect worth knowing about

The final whole-branch review found that **the exit animation never played on the first dismissal of any session.** The latch gating it was a passive `useEffect`, which Preact defers past a paint — so the first commit rendered the panel `display: none`, and a CSS transition cannot start from `display: none`. It jumped straight to the end state.

It was invisible to 930 passing tests *and* to a 27-check browser pass, because every page in that harness dismissed exactly once — so every check only ever exercised the broken path. The reviewer found it by reading Preact's scheduler.

The corrected harness now measures it directly:

```
first dismissal   opacity 0.571604   (was 0 — jumped)
second dismissal  opacity 0.566085
```

## Verification

939 tests, coverage 81.33% (floor 74.3), lint silent, build green on Node 24.

Browser, 390×844: **29/29** on the panel and **35/35** on the day rail. Includes the two defects an earlier review round found that no unit test can see — a bubbling `transitionend` from a descendant cutting the slide short, and the dismissal's own correction knocking out the IntersectionObserver sentinel that gates its own animation.

## Known, deliberately parked

- `page.tsx` is rendered by no test (pre-existing), and the browser harness is not in CI — so a refactor of the consumer half could reintroduce the no-animation bug with all tests green.
- `finish()` doesn't reset `exitScrolledPast` the way `cancelExit` does. Harmless today; every consumer gates on `exiting` first.
- The /about guide mentions scrolling but not that a desktop click outside also dismisses.
- `⟳ Now` keeps its smaller tap target while the Filters toggle gained a 44px one.

Nothing under `ios/` is touched.
[skip-screenshots: web-only change, no iOS pixels touched]

https://claude.ai/code/session_01XMxQocbmXTw5K2fF4RPmb2


bbernstein:
## Chrome could not scroll this page slowly — fixed in `cf97804`

Found on localhost, then **reproduced identically on production**, where it has been live since #238. Slow scrolling advanced the page ~120px and snapped back to the top, over and over.

**It is a feedback loop that the browser's scroll anchoring closes:**

1. ~64px of scrolling took the sentinel above the viewport → the filter card was hidden
2. hiding it removed **~290px of in-flow height above the reader**
3. scroll anchoring held the reader's event card still by subtracting that from `scrollY` — which clamped at the top of the document
4. back at the top the sentinel was visible again → the card returned → repeat

Measured: **40 slow wheel ticks (2400px requested) advanced the page 0px and returned to the top 20 times.** Instrumenting every scroll API *before* app scripts ran recorded **zero programmatic scrolls**, so this is the browser compensating, not our correction misfiring. Chromium and WebKit both reproduce it.

**The root cause is geometric, not a browser quirk: a ~290px header cannot collapse after 64px of scrolling.** There is no room above the reader to absorb it, so the collapse destroys its own precondition. `overflow-anchor: none` would suppress the symptom at the cost of a measured ~292px content jump, so it was rejected.

### The fix

The card is never taken out of flow. The sticky header rides up by exactly the card's measured height (`--filter-card-h`) and pins there, parking the card just above the viewport with the rail flush at the top edge. Document height never changes, anchoring has nothing to correct, and the card scrolls away under its own steam — which is what "behave the way it looks" was asking for in the first place.

| | before | after |
|---|---|---|
| 1800px of slow scrolling delivers | **0px**, 15 of 30 ticks backwards | **1800px**, 0 backwards |
| parked card | `display: none` | in flow, `inert`, `bottom: -16` |
| rail | — | `top: 0`, flush |

Three consequences, none optional:

- **`inert` while parked.** Still in the DOM and still in flow, so without it a keyboard reader Tabs into it and the browser scrolls the header back down to show them the focused control. `display: none` gave this for free.
- **The sentinel moved below the header.** Above it, the signal fired 64px in while the card was still half on screen — and the exit animation briefly takes the card out of flow, so only a sentinel *below* moves by the same amount anchoring subtracts, keeping the signal stable under its own animation.
- **The exit's end needed a scroll correction.** Every close path already captured a reference before the panel *left* flow; nothing captured one for its *return*, which never mattered while it returned to `display: none`. It now genuinely re-enters flow — measured as a **285px jump down the list at the exact moment the animation ended**, caught by check 13.

### Verification

954 tests (15 new, each mutation-tested — I broke all three mechanisms and confirmed the tests fail), coverage 81.41% against a 74.3 floor, lint silent, build green.

Browser at 390×844: **34/34** on the panel and **35/35** on the rail. Five of the panel checks are new and specific to this bug; **they were run against production as a control and fail there**, on the live bug. One of them passed spuriously at first — its fallback selector found the wrong element — and now identifies the card the way the accessibility tree does, via the toggle's `aria-controls`.

PR #241, opened 2026-08-17
fix(web): make the day rail's highlight move with your finger

bbernstein:
## The problem

The rail's highlight arrived in two discrete jumps that disagreed with each other:

- `useDayAnchor` emitted a **day key**, flipping the instant a section's top crossed the sticky line.
- `DayRail`'s effect fired on `[anchorDay]` and teleported `strip.scrollLeft` to centre that chip.
- Meanwhile a CSS `transition-colors` crossfaded the colour on its own timer.

Reported as: *"the highlight changes to the other date and then moves to the centre location."*

## The mechanism

The day-title replacement in the list is **scroll-linked, not time-linked** — sticky headers handing over across a fixed distance of scroll, which is why they track your finger and stop half-done when you stop. The rail now works the same way: the current day is a *fraction*, ramped over `max(headerHeight, min(25% of the section, 200px))` at the day's trailing edge. Crisp while reading, visibly continuous at flick speed.

**The fraction never becomes React state.** `page.tsx` is the whole application, so a per-frame `useState` would re-render it 60×/second. Everything is measured in a rAF and written to three DOM properties. Zero re-renders while scrolling.

## Interaction model

| Gesture | Page | Rail `scrollLeft` | Pill |
|---|---|---|---|
| Scroll the list | moves | follows, keeping the pill centred | tracks `f`, glides day→day |
| Drag the rail sideways | **still** | reader owns it; sync suspended | stays glued to its day, drifts off-centre |
| Tap a chip / `‹` `›` / `⟳ Now` | scrolls | resumes, tweens ~220ms | eases to centre |
| Scroll into a new day after a peek | moves | sync resumes here | re-centres |

**Dragging a day into the centre does not make it current.** The pill isn't a selector parked at the centre — it's attached to the day being read, and only *appears* centred because the rail auto-scrolls to keep it there. Drag peeks; tap commits. That keeps every commit an explicit button activation, so assistive tech still hears "Go to Saturday, July 4, 12 events" on every navigation — a release-to-commit drag announces nothing.

The invariant behind all of it: **`anchorDay` is derived from page scroll and nothing else.** The rail only ever reads it.

## Painting

Three layers in one scroller, so they share a single `scrollLeft` and cannot desync. The pill sits **above** the chips (below it, a chip's `hover:bg-blue-50` paints over the highlight on the current day), and a clipped copy of the row in the highlighted colour sits above the pill. A chip straddling the edge is genuinely split — left half base colour, right half white.

## Browser verification

390×844 Chromium against real production data. **This is where the one real defect was found — the branch was green on all 1029 tests at the time.**

Suspending on `wheel`/`pointerdown` over the strip infers intent from event type, and that inference is wrong for the most ordinary gesture there is: a plain **vertical** page scroll with a finger resting on the rail — which, since the rail is sticky at the top of a phone, is where swipes start. Measured: one vertical wheel, then 120px of page scroll, moved the strip **0px**; centring stayed dead until the next day boundary. Replaced with divergence detection (every write reads its value back; a strip scroll landing somewhere we didn't put it means the reader moved it). Same measurement after: **24px**. Strictly more accurate in both directions, and it now covers scrollbar drags and keyboard too.

Also measured:

- **Layer alignment** across all 251 chip pairs: **0px** worst delta.
- **Cost of the duplicated row**: none. Median frame 8.3ms with the copy, 8.3ms with it `display:none` (p95 9.9 vs 10.0). The rail turned out to span **251 chips**, not the ~64 the design estimated, so the copy is 502 buttons over a 12044px strip.
- **`clip-path` coordinate space**: resolves against the layer's own border box. Mid-handover `inset(0 1079.88px 0 10920.1px)` on a 12044px box; 10920.1 + 44 + 1079.88 = 12043.98.
- **Continuity**: 13 distinct pill positions across the 200px ramp, and the last pre-flip position *equals* the post-flip one (10896 + 48 = 10944) — continuous **across** the anchor change, not merely smooth either side of it.
- **Commit**: tapping six days ahead eases the strip through 24 distinct positions over ~400ms; under `prefers-reduced-motion`, 2.

### Not verified

Native axis-locking on a diagonal touch drag starting on the rail — headless Chromium doesn't arbitrate real touch gestures, so this needs a device or simulator pass. Low risk (ordinary `overflow-x-auto`, no `touch-action` override), but it's untested rather than passing.

## Tests

1029 passing. 29 pure geometry cases, 15 hook cases against stated layout, 4 component guards on the copy being paint-only — **each falsified by breaking the code**, including one guard I had to rewrite because it filtered on `[data-chip]`, which the copy doesn't carry, and so could never have failed.

One existing `DayRail` test was re-pointed at the new trigger rather than retired: it guards "never `scrollIntoView`", which didn't change.

## Notes for review

- `resolveAnchor` is now shared by `useDayAnchor` and the highlight. That's load-bearing, not tidiness: the pill and `aria-current` must never name different days, and one walk cannot disagree with itself. `DayRail` takes `windowDayKeys` for the same reason — both walks get identical input rather than input that happens to agree.
- Adds `DAY_HEADER_ATTR` so the ramp's floor reads a declared element rather than `firstElementChild`.
- Design and findings: `docs/plans/2026-08-17-day-rail-scroll-linked-highlight.md`.

[skip-screenshots: web-only change, touches no ios/ paths]

https://claude.ai/code/session_01AyVVFZYXXcmS6RrqtA2igP

### DIFF
--- a/docs/plans/2026-08-17-day-rail-scroll-linked-highlight.md
+++ b/docs/plans/2026-08-17-day-rail-scroll-linked-highlight.md
@@ -1,7 +1,7 @@
 # Day rail — scroll-linked highlight
 
-**Status:** Implemented and browser-verified. Branch pushed, PR not yet
-opened.
+**Status:** Implemented and browser-verified on branch. Not pushed, no PR
+yet.
 **Branch:** `feat/web-rail-scroll-linked-highlight`
 **Follows:** phase 3a (#238, #239, #240) — the day rail itself.
 
@@ -96,9 +96,11 @@ any explicit commit. Snapping back to centre the instant the reader scrolls
 one pixel would destroy the peek they just made; scrolling around within
 today's events keeps it, scrolling into tomorrow re-centres.
 
-**Suspension trigger:** `pointerdown` **and** `wheel` on the strip. A
-trackpad two-finger horizontal pan fires no pointer event and would otherwise
-be fought by the per-frame writes.
+**Suspension trigger:** the strip's scroll position diverging from the last
+value this hook wrote. *(Revised during the browser pass — the original plan
+said `pointerdown` and `wheel` on the strip, which also fires for a plain
+vertical page scroll that merely passed over the rail. See the browser
+verification section below.)*
 
 ## Architecture
 
@@ -142,17 +144,31 @@ thrash.
 ```
 <div ref=strip class="overflow-x-auto">                              scroller
   <div class="relative flex gap-1 w-max">                            content
-    <div ref=pill  class="absolute bg-blue-600 rounded-md" />        z-0
-    {chips}                                                          z-10  base colours
+    {chips}                                                          static, base colours
+    <div ref=pill  class="absolute inset-y-0 bg-blue-600" />         z-10
     <div ref=clip aria-hidden class="absolute inset-0 flex gap-1
          text-white pointer-events-none" style="clip-path: inset()"> z-20  duplicate row
-      {chips as divs, not buttons}
+      {chips again, as buttons with tabIndex -1}
     </div>
   </div>
 </div>
 ```
 
 Both layers share one `scrollLeft`, so they cannot desync.
+
+Two details settled while building, against the sketch above:
+
+- **The pill sits ABOVE the chips, not behind them.** Behind, a chip's
+  `hover:bg-blue-50` paints over the highlight on the current day — the one
+  chip whose highlight matters. Above, the pill covers the base text and the
+  clipped copy puts legible text back, which is the same result the layering
+  was for.
+- **The copy's chips are `<button>`, not `<div>`.** The copy must match the
+  real row's box metrics exactly or a width difference opens a seam through a
+  digit, and a `<div>` does not inherit the same UA and preflight rules.
+  `tabIndex={-1}` inside an `aria-hidden` container keeps them out of the tab
+  order and the accessibility tree. Measured worst delta across 251 pairs:
+  0px.
 
 **The pill and the clip layer live in content coordinates**, computed from
 `f` alone and never touched by `scrollLeft`. `scrollLeft` is a wholly
--- a/docs/plans/2026-08-17-day-rail-scroll-linked-highlight.md
+++ b/docs/plans/2026-08-17-day-rail-scroll-linked-highlight.md
@@ -147,7 +147,7 @@ thrash.
     <div ref=pill  class="absolute inset-y-0 bg-blue-600" />         z-10
     <div ref=clip aria-hidden class="absolute inset-0 flex gap-1
          text-white pointer-events-none" style="clip-path: inset()"> z-20  duplicate row
-      {chips again, as buttons with tabIndex -1}
+      {chips again, as non-interactive divs}
     </div>
   </div>
 </div>
@@ -162,12 +162,16 @@ Two details settled while building, against the sketch above:
   chip whose highlight matters. Above, the pill covers the base text and the
   clipped copy puts legible text back, which is the same result the layering
   was for.
-- **The copy's chips are `<button>`, not `<div>`.** The copy must match the
-  real row's box metrics exactly or a width difference opens a seam through a
-  digit, and a `<div>` does not inherit the same UA and preflight rules.
-  `tabIndex={-1}` inside an `aria-hidden` container keeps them out of the tab
-  order and the accessibility tree. Measured worst delta across 251 pairs:
-  0px.
+- **The copy's chips are `<div>`, not `<button>`.** They were buttons first,
+  chosen for box-metric parity, and `e2e/verify-rail.mjs` proved that wrong:
+  the rail's own chevron selector is `button:not([data-chip])`, the copy
+  carries no `data-chip`, so the copy matched it and a chevron click landed on
+  inert paint. Adding `data-chip` to fix that would instead double every
+  `[data-day-rail] [data-chip]` query in the same suite; a `<div>` is the only
+  shape satisfying both. Parity turned out not to need a button at all —
+  Tailwind's preflight normalises a button's font, padding, border and margin
+  to a div's. Measured worst delta across all 251 pairs: 0px on left, width,
+  top and height.
 
 **The pill and the clip layer live in content coordinates**, computed from
 `f` alone and never touched by `scrollLeft`. `scrollLeft` is a wholly
@@ -201,10 +205,16 @@ reader's own gesture and stays as-is.
 - **Pure** (`railPosition`): ramp floor on a one-event day, clamping at both
   ends of the season, `f` continuity across the flip.
 - **Hook** (`useRailHighlight`): stubbed rects, following the existing
-  `useDayAnchor.test.ts` pattern. Suspension on `pointerdown`/`wheel`; resume
-  on anchor change.
+  `useDayAnchor.test.ts` pattern. Suspension by scroll divergence — the strip
+  landing somewhere other than the last value written — and *not* on a
+  vertical scroll that merely passed over the rail; resume on anchor change,
+  on an explicit commit, and against a scroll event still in flight from the
+  pan just lifted. Plus listeners attaching to elements that appear after the
+  first mount, and no listeners at all while nothing is rendered.
 - **Component** (`DayRail`): clip layer is `aria-hidden`, contains nothing
-  focusable, `aria-current` still lands on the anchor.
+  focusable, contributes nothing to the chevron selector, `aria-current`
+  still lands on the anchor, and the highlight attaches when the strip
+  appears only after the scope resolves.
 - **Browser, mandatory.** Chrome on a phone viewport, with a GIF of a slow
   drag *stopped mid-transition* showing the half-and-half chip. This is the
   acceptance criterion and no test can reach it — phase 3a shipped eleven
--- a/docs/plans/2026-08-17-day-rail-scroll-linked-highlight.md
+++ b/docs/plans/2026-08-17-day-rail-scroll-linked-highlight.md
@@ -1,7 +1,6 @@
 # Day rail — scroll-linked highlight
 
-**Status:** Implemented and browser-verified on branch. Not pushed, no PR
-yet.
+**Status:** Implemented and browser-verified. In review as PR #241.
 **Branch:** `feat/web-rail-scroll-linked-highlight`
 **Follows:** phase 3a (#238, #239, #240) — the day rail itself.
 
--- a/docs/plans/2026-08-17-day-rail-scroll-linked-highlight.md
+++ b/docs/plans/2026-08-17-day-rail-scroll-linked-highlight.md
@@ -1,6 +1,7 @@
 # Day rail — scroll-linked highlight
 
-**Status:** Approved design, not yet implemented.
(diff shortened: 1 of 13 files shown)


bbernstein:
## Review responses

All findings from both reviewers are addressed. Neither review left inline threads, so **there is nothing to mark resolved** — Copilot's findings arrived inside a collapsed `Suppressed comments` block on its review body, and Claude's as an issue comment.

### Copilot

Worth flagging for anyone reading this later: the review says *"generated no new comments"* and then hides **three real findings** in the collapsed `Suppressed comments` block. All three were valid.

| Finding | Status |
|---|---|
| `installLayout` stubs `Element.prototype.getBoundingClientRect` without restoring it — leaks across test files | Fixed (`96ed0b9`) |
| ...and captures the *already-stubbed* version as its fallback, so stubs stack per `mount()` | Fixed — the real implementation is captured once at module load, restored in `afterEach` |
| Plan doc says "Not pushed, no PR yet" while being added in this PR | Fixed |

The stacking half was the sharper catch: seventeen mounts nested seventeen stubs, each delegating to the last.

### Claude

**1. Comments claiming the copy carries `data-chip` — fixed (`ef6f3bf`).** This was more than stale prose. The copy carrying no `data-chip` is *exactly* what broke `e2e/verify-rail.mjs`: its chevron selector is `button:not([data-chip])`, the copy's chips were `<button>`, so the copy matched it and a chevron click landed on inert paint with the real chip intercepting beneath. The comments described the code as safe in the precise way it had just failed to be.

Fixed at the root rather than in the comment. The copy's chips are now `<div>` — the only shape satisfying both selector families the e2e suite already uses:

```
<button>, no data-chip  -> matches the chevron selector   (the bug)
<button> + data-chip    -> doubles every [data-chip] query
<div>                   -> correct for both
```

`<button>` was originally chosen for box-metric parity, which turns out not to require it — Tailwind's preflight normalises font, padding, border and margin. Re-measured: **0px** worst delta across all 251 pairs on left/width/top/height, rail now reports 4 non-chip buttons rather than 255, `[data-chip]` still returns 251 rather than 502.

**2. Listener effect churn — fixed, then re-fixed.** The first attempt made the effect mount-scoped, and that was wrong in a way worth recording: `DayRail` renders nothing until events load, so the elements don't exist on first mount. The effect attached neither the strip's `scroll` listener nor the content `ResizeObserver`, and never re-ran. **Peek detection would have been dead in production.** Every existing test mounts with chips already present, so none could see it.

Now keyed on the elements themselves via callback refs and a generation counter — re-runs exactly when a node attaches, once on load. The original churn complaint stays fixed. Regression test mounts empty and re-renders with chips, and fails against the mount-scoped version.

**3. Ordering assumption behind divergence detection — documented.** Spelled out on `lastWritten`: a `scroll` event is dispatched no earlier than the task that caused it and reads the live position, so our own writes always compare equal, even mid-tween where a write happens every frame.

### Also fixed, found by re-reading the diff

`resume()` cleared the suspend flag but left `lastWritten` pointing at where *we* last wrote, not where the reader left the strip — so a `scroll` event still in flight from their pan compared divergent and re-suspended immediately, silently undoing the resume. Both `resume` and the anchor-change path now re-baseline. Tested, and falsified by removing it.

### On the `browser-checks` failure in the previous run

`11c ⟳ Now hides once back on today` failed at 35/36. I do not believe it is this branch:

- Passes locally 3/3.
- Measured headroom: `⟳ Now` hides in **138ms** unthrottled and **543ms at 6× CPU throttling**, against the test's 1500ms budget.
- `main`'s own HEAD (`f9e8a31`) failed this same suite 35/36 on a *different* check (`12 header clears the rail @ 200% zoom`), so the suite flakes on CI independently of this work.

Watching the current run rather than asserting it's fine. The earlier `browser-checks` failure in the run before that was a genuine bug in this branch and is fixed.

### Still not verified

Native axis-locking on a diagonal touch drag starting on the rail — headless Chromium doesn't arbitrate real touch gestures, so it needs a device or simulator pass. Untested rather than passing.

Every new guard in this PR was proved by breaking the code first.

bbernstein:
## `browser-checks` — root cause found, and it is not this branch

I called `11c ⟳ Now hides once back on today` a flake earlier. **That was wrong, and I want to correct it plainly.** It failed all three times CI reached it. It is deterministic, and it is reproducible — I just had the wrong variable.

**It is the timezone.** CI runs UTC; I had only ever run locally in Eastern. Clean A/B, same machine, same minute, built `vite preview` in both cases:

| Branch | Timezone | `11c` |
|---|---|---|
| PR #241 | Eastern | **PASS** (11 runs) |
| PR #241 | UTC | **FAIL** |
| `main` | Eastern | **PASS** |
| `main` (worktree, `f9e8a31`) | UTC | **FAIL** |

**`main` fails identically.** This PR is not the cause.

### Why UTC breaks it

The app's default scope is `Now` — events from the current time forward. Event datetimes are Chautauqua-local and naive, so under a UTC browser the app effectively believes the local clock *is* event-time. At 21:35 UTC it thinks it is 9:35pm at the Institution, by which point the day's programming has ended, so today retains no upcoming events. `⟳ Now` then cannot land the anchor on today, and the button correctly stays visible — the assertion is what is wrong, not the app.

That predicts the observed history exactly, and it holds:

- `main`'s passing run: **19:36 UTC** (app thinks 7:36pm — evening programme still running).
- This branch's failures: **21:22 and 21:31 UTC** (app thinks 9:22/9:31pm — day over).

So **`browser-checks` fails for everyone, on every branch, after roughly 20:00 UTC**, and passes earlier in the day. It is a wall-clock-dependent check. It has run on `main` exactly once since being introduced in #240 — that run was before the cutoff.

### What I did not do

I have not touched `e2e/verify-rail.mjs`. It is not this PR's test, the failure is pre-existing, and "fix the failing test" is exactly the move that should be asked about rather than performed. Two options for whoever picks this up, as a separate change:

1. **Pin the browser timezone** in `newPage` (`timezoneId: 'America/New_York'`), matching the domain the data lives in. Smallest fix, makes the suite deterministic, and arguably what every date-sensitive check here wants.
2. **Make `11c` time-independent** — assert against a day that has events under the current scope rather than assuming today does.

Option 1 also removes a latent trap from every other date-sensitive check in the file.

### Separately: is the app's UTC behaviour itself a bug?

Possibly, and worth its own issue rather than a fix smuggled in here. A visitor browsing from a UTC (or any non-Eastern) device late in their local day sees today's Chautauqua events as already past. That is a real user-facing question about whether "now" should be evaluated in the Institution's timezone rather than the device's — out of scope for a rail-highlight PR, but I would rather flag it than leave it found-and-forgotten.

---

## Latest review round

Both of Copilot's newest suppressed findings addressed in `08255da`:

- **Layout effect re-measured the chip row on `keysId`** — valid. The row's geometry depends on the chips, never on the day window, so every filter and scope change forced a layout plus 251 `getBoundingClientRect` reads for a row that had not moved. Split into measure-on-chips and place-on-window, with declaration order as the ordering contract.
- **Duplicate scroll-spy walk with `useDayAnchor`** — valid observation, measured rather than acted on. 0.034ms per walk at a typical render window (3 mounted sections), 0.335ms at a large one (16 sections, 321 cards) — both hooks together are 0.4% of a frame typically, 4% at the large end. Sharing one measurement means threading a subscription through `page.tsx` or hoisting a third hook; that is a new coupling across a boundary this design keeps clean, bought with a measured 4% worst case. Left as is, with the number recorded at the call site so the next reader has it.

### One earlier finding rejected

Copilot asked for `MutableRefObject` instead of `RefObject`, on the grounds that `RefObject.current` is read-only. `@types/react` here is **19.1.8**, where `RefObject<T>` is `{ current: T }` — mutable — and `MutableRefObject` carries `@deprecated Use RefObject instead`. The change would swap correct code for a deprecated alias to fix a violation that does not exist. No change made.

bbernstein:
## All review comments have been addressed

**Threads needing manual resolution: none.** Both reviewers commented at the review-body level rather than opening inline threads, so there are zero review threads on this PR — nothing to mark resolved.

### All checks passing

| Check | |
|---|---|
| `browser-checks` | ✅ pass |
| `claude-review` | ✅ pass |
| `screenshot-freshness` | ✅ pass |
| `test-backend (24)` / `(25)` | ✅ pass |
| `test-frontend (24)` / `(25)` | ✅ pass |

`mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, `pendingReviewers=0`.

`browser-checks` has now gone green twice consecutively at ~22:00 UTC — the window in which it failed on *every* branch, including `main`, before the timezone pin.

### Findings across 8 review passes: 14, all resolved

| Source | Finding | Outcome |
|---|---|---|
| Claude | Stale `data-chip` comments | Fixed — same root cause as the first CI failure |
| Claude | Listener effect churn on every filter change | Fixed, then re-fixed (see below) |
| Claude | Divergence ordering assumption undocumented | Documented |
| Claude | Mount-scoped effect misses late-arriving DOM | Fixed via node generation |
| Copilot | `getBoundingClientRect` stub leaks and stacks per mount | Fixed |
| Copilot | Plan doc status stale | Fixed |
| Copilot | Global listeners attached while nothing rendered | Fixed |
| Copilot | Plan doc ×3: DOM sketch, copy element type, suspension mechanism | Fixed |
| Copilot | Layout effect re-measures 251 chips on `keysId` | Fixed — split the effect |
| Copilot | Duplicate scroll-spy walk | Measured, not acted on; number recorded at the call site |
| Copilot | `lerp` doesn't land exactly on `t=1` | Fixed — **and my comment was backwards** |
| Copilot | Use `MutableRefObject` | **Rejected** — deprecated in `@types/react` 19.1.8 |

**Every new guard was proved by breaking the code first.** Two of the guards I wrote turned out to be can't-fail tests and were replaced: one filtered on `[data-chip]`, which the copy doesn't carry; the other used integer coordinates, for which the unclamped `lerp` happens to be exact.

### Three bugs no unit test could have caught

1. **Peek detection inferred intent from event type** — killed centring on any vertical scroll passing over the rail. Measured 0px vs 24px.
2. **The copy rendered as `<button>`** — collided with the e2e's chevron selector.
3. **Two null-ref effect bugs I introduced while fixing review feedback** — the second found by the test written for the first. Peek detection would have been dead in production.

### `browser-checks` — root-caused, not worked around

It was never a flake. CI runs UTC; event `startDate`s are Institution-local and naive, so the app treated the runner's clock as event-time and believed the day's programming was over. `main` failed identically. Fixed by pinning `timezoneId: 'America/New_York'` on both suites — which also de-risks every other date-sensitive check in those files.

Note for whoever merges: this pin lands on `main` with this PR, so no cherry-pick is needed.

### Two things that outlive this PR

1. **A product question, worth its own issue.** The timezone work exposed it but did not answer it: a visitor on a non-Eastern device late in their local day sees today's Chautauqua events as already past, because "now" comes from the device clock while event times are Institution-local. I pinned the *test's* clock only — deliberately not the app's, inside a CI commit.
2. **Native touch axis-locking on a diagonal drag starting on the rail is unverified.** Headless Chromium doesn't arbitrate real touch gestures. Low risk (ordinary `overflow-x-auto`, no `touch-action` override), but untested rather than passing — worth a pass on a real device.

### Ready for merge

Not merging — leaving that to @bbernstein.

PR #245, opened 2026-08-19
feat(ios): day rail on the events list (phase 3b)

bbernstein:
Phase 3b of the cross-platform date navigation initiative (#226): the iOS Events tab gets the same day rail the web shipped in phase 3a (#238/#241).

A horizontal strip of day chips pinned above the list, spanning the whole navigable season. Tap a day to go there, step by one day with the chevrons, or ⟳ Now back to today. The highlight follows your scroll, so the rail answers "where am I?" rather than "what did I last tap?".

Spec: `docs/superpowers/specs/2026-08-15-cross-platform-date-navigation-design.md`
Plan: `docs/superpowers/plans/2026-08-18-date-navigation-phase-3b-ios-day-rail.md`

## Reviewer attention — three behaviour changes

**1. `expandWindowEnd()` steps to the next day that HAS events, not the next calendar day.**
Previously it widened by one calendar day. With Favourites on, or any search that leaves gaps, the adjacent day usually has no matches — the edge moved, nothing mounted, and the control read as dead. This is the correction the web already shipped. It invalidated `expandWindowEndSetsWindowEndDayKeyOneDayAtATime`, whose fixture put all 50 events on one day; the fixture gained events on 08-06/08-09 and the test now pins the new rule *and its point* (that 08-04 and 08-05, being empty, are skipped).

**2. `⟳ Now` is navigation-only; `AppModel.resetToNow()` is deleted.**
It used to reset the scope, clear `selectedWeeks`, and persist — so a reader with a week filter silently lost it by tapping Now. The spec is explicit and platform-neutral: ⟳ Now "does not touch scope, weeks, categories, or search… Keeping it filter-free is the premise of the whole design", and web's equivalent is `goToDay(todayKey)` and nothing else. Two Task 6 tests were rewritten to the new contract and one added pinning that a week selection survives Now. The `DateFilterSheet` scope buttons remain how you reset scope and window.

**3. `AppModel.expandWindowStart()` deleted** — it had no production caller; the backward chevron routes through `goToDay`.

## The defect twelve green task reviews missed

Tapping a chip for a day **already inside the window** did nothing, and left an armed pending scroll that then froze the rail's highlight for the rest of the session. The resolver only ran from `.onChange(of: days.map(\.id))` and `.onAppear`, and an in-window target changes neither. Every positive-path test tapped a day *outside* the window, where expansion changes `days` and hides the gap.

Found by the whole-branch review, confirmed on a running app, fixed with `.onChange(of: pendingScroll)`, and now pinned by two tests that fail without it. Same shape as the lesson phase 3a paid for.

## New test infrastructure

This repo had **no UI-test target** — `capture-screenshots.sh` drives DEBUG launch hooks through `simctl` precisely to avoid needing one. But screenshots cannot prove a rail stays pinned while you scroll, that a tap does not drag the list, or that a distant tap lands on its target: the three defects phase 3a's browser pass caught after eleven clean reviews.

- `ChqCalendarUITests` (13 tests), wired into `ios.yml` in the same invocation; `timeout-minutes` 45→60.
- `-uitest-fixture`: a DEBUG-only deterministic data path (generated in Swift, **not** a bundled JSON — synchronized folder groups would ship a resource into Release builds; `#if DEBUG` excludes code, not resources), with an in-memory cache and a throwaway `UserDefaults` suite so a fixture launch cannot poison a real one.

Every UI test pins both data and time. CI runs UTC and this app treats the device clock as Chautauqua event-time.

## Tests

- Unit: `✔ Test run with 846 tests in 62 suites passed`
- UI: `Executed 15 tests, with 0 failures` in one full-target run from a shut-down simulator
- Screenshots regenerated with the real scripts; shots 01–03 now show the rail. Also fixed `03-search` capturing blank against live 1,686-event data (settle 6s→12s).

## Known follow-ups (not in this PR)

- **Accessibility, and this one gates 1.1.3 submission rather than merge:** an on-device `performAccessibilityAudit()` found 18 contrast failures and 21 Dynamic-Type warnings on the chips at the OS's *default* text size. `DayChip` is byte-equivalent to the `MyDayChip` it replaced — no new contrast defect — but the rail promotes it to ~64 instances on the primary screen, and this PR regenerates the App Store screenshots, so as-styled is what review sees. The fix is a styling pass that must also cover My Day.
- ⟳ Now scrolls with the strip rather than being screen-pinned.
- VoiceOver verified by accessibility-tree dump, not a live audio pass.
- After a rail tap the highlight lands one day short (pre-existing `visibleDays` precision under programmatic `scrollTo`).
- The rail's ~64 chips sit in a non-lazy `HStack`.
- Deliberately out of scope per the plan: Siri routing, the 1.1.3 version sweep, listing-copy review, removing `DateScope.thisWeek`, `stepWeek`.

https://claude.ai/code/session_01Dxr9zVJLDHofJBEs26bUw4


bbernstein:
## Accessibility follow-up: day chip contrast (2 commits)

Addresses the accessibility finding recorded in this PR's description. Three defects were real; a fourth turned out to be a measurement artifact, and that distinction shaped the fix.

### The real defects

1. **`.disabled()` dims a control's entire label as a compositing pass applied from outside it.** The Events rail disables empty-day chips, so their text rendered at a *measured* ~3.7:1 no matter what `foreground` asked for. An empty chip is now a **non-interactive view** rather than a disabled `Button` — same affordance, same factual accessibility label ("Sunday, August 16, no events", never "Go to"), no dimming pass. My Day's chips are untouched: it never disables them, so they remain buttons and its empty chips stay tappable for the "Browse …" flow (#192).
2. **The selected chip was white on the accent `#5B7F95` = 4.27:1**, just under WCAG AA. New chip-local `DayChipSelected` asset (light `#3E5868`, dark `#4D6C7F`). The global `AccentColor` is deliberately unchanged.
3. **`.thinMaterial` made contrast depend on whatever scrolled behind the chip.** Now an opaque `DayChipBackground` (light `#F2F2F7`, dark `#1C1C1E`), so contrast is deterministic rather than a function of scroll position.

### Why contrast is gated by computation, not by the audit

`performAccessibilityAudit` cannot sample individual text runs once a `Button` collapses them into one accessibility node, so it reports contrast failures against chip text whose real pixels are **18.8:1** (light) and **17.0:1** (dark). The only structure that satisfies it — a hidden visual layer under an invisible `Color.clear` button — broke tap delivery after a drag and cost 6 UI tests. Scroll-the-rail-then-tap is the rail's primary interaction, so that trade was rejected rather than shipped.

Instead, `DayChipContrastTests` resolves the committed asset colours in **both** appearances and computes WCAG 2.1 ratios directly:

| | ratio | AA (4.5:1) |
|---|---|---|
| Light, unselected | 18.82:1 | pass |
| Dark, unselected | 17.01:1 | pass |
| Light, selected | 7.50:1 | pass |
| Dark, selected | 5.57:1 | pass |
| *old accent, selected* | *4.27:1* | *fail* |

That last row is a built-in self-check that the helper can detect a failure; the asset path was falsified separately by pointing `DayChipSelected` at the old accent and watching the test go red.

The scoped UI audit is retained for what it *can* measure on the rail — text clipping and Dynamic Type on the individually-identified controls.

### Known limitation, stated rather than hidden

The audit's Button-collapse blind spot also produces a false Dynamic Type finding on a chip's count digit, so those findings are dropped when the element has no identifier of its own. That costs real coverage: a future regression introducing genuinely broken text *inside* a chip's collapsed label would present the same shape and go uncaught. Documented in the test file.

### Also in these commits

- The rail's hardcoded `44×62` control frames become a 44pt *minimum* that can grow, so text stops clipping at large Dynamic Type sizes while the tap target stays HIG-legal.
- The blank `Text(" ")` count placeholder is replaced by a `@ScaledMetric` height reservation, so chip heights stay stable at every text size.
- Screenshots regenerated — the chips changed appearance, and My Day's changed with them, since both screens share `DayChip`.

### Verification

- `✔ Test run with 851 tests in 63 suites passed` (846 baseline + 5 new)
- Whole `ChqCalendarUITests` target green in one run: 17 tests, 0 failures
- Checked by eye in light and dark: the selected chip still reads as selected, and the empty-day dashed treatment is legible now its text is undimmed

bbernstein:
## Review iteration complete

All review comments have been addressed. Copilot's most recent pass (2026-08-20T00:13Z) is the first with **no `Suppressed comments` block** — the previous four passes each reported "generated no new comments" while carrying real findings in that collapsed section, so the empty block is the actual signal here.

### Findings addressed

| Finding | Resolution | Commit |
|---|---|---|
| `railBounds(in:)` unioned `app.buttons` only, so the audit skipped empty chips — which are no longer buttons | Bounds now union all descendants matching the chip/control identifiers. Widening surfaced 0 new issues, but the gate now actually covers the chips whose dimming was the original defect | `71b95c9` |
| Two audit tests named for "Contrast" while `auditTypes` excludes `.contrast` | Renamed; the doc comments explaining the contrast/audit split are kept | `71b95c9` |
| `Foundation.framework` pinned to `iPhoneOS26.0.sdk` | Reference, build file, phase entry and the empty group wrapper removed; target builds and runs without it | `71b95c9` |
| `rebuildDerivedCounts()` re-ran a full filter pass on window-only changes | Skipped unless snapshot, favourites, or the non-window filter identity changed. Pinned by a rebuild counter, so the test asserts non-recomputation rather than merely unchanged output | `71b95c9` |
| Spec still documented the rail's end controls as scrolling "by design" | Corrected — that note would have taught the next reader that the pinned controls are the regression | `aabf9ba` |
| Both disabled chevrons announced the same "No further days in this direction" | Each names its own direction; at a season edge both are disabled at once, so identical labels left a VoiceOver user unable to tell earlier from later | `9f34164` |
| `navMatching` fingerprint includes `dateScope`/`selectedDayKey`, which the rebuild overwrites | **Answered, not changed.** The waste is one imperceptible pass per deliberate scope tap; the hot path (`expandWindowEnd` during scroll) is already skipped. A narrower fingerprint would save that pass and risk silent staleness if it ever omits an input that matters. Recorded in the code | `5c6dc02` |

### Threads to resolve manually

- `ios/ChqCalendar.xcodeproj/project.pbxproj:53` (Copilot, `Foundation.framework`) — fixed in `71b95c9`; the reference and its build-phase entries are gone.

### Checks

All 8 required checks pass on `5c6dc02` — iOS (852 unit + 18 UI, serial on a 3-core runner), browser-checks, frontend ×2, backend ×2, screenshot-freshness, claude-review. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, no pending reviewers.

**Ready for merge.** Not merging — that's the author's call.

PR #248, opened 2026-08-20
fix: local dev — pin LocalStack, create tables on every start, repair stale scripts

### LINKED ISSUE #247: Local dev environment: four defects that let `setup-local.sh` report success on a broken stack
# Local dev environment: four defects that let `setup-local.sh` report success on a broken stack

## Summary

A from-scratch verification pass on a clean `main` (`afbee1b`) found four independent defects in local dev setup. Every automated suite is green — backend 909, frontend 1099, iOS 789, publisher-format 18, plus `terraform fmt`/`validate` — so none of this is caught by CI.

They share a theme: **`./scripts/setup-local.sh` exits 0 while leaving the environment unusable.** That is the same failure mode #214 set out to eliminate, so these look like remaining gaps in that work rather than new regressions.

Environment: macOS 26 (arm64), Docker 29.6.1 via Colima, Node 24.18.0, npm 11.16.0.

---

## 1. LocalStack is license-gated on `:latest` and exits immediately

`docker-compose.yml:91` pins `localstack/localstack:latest`. That tag now resolves to LocalStack **2026.7.5**, which requires an auth token and quits on startup:

```
chq-calendar-localstack  | LocalStack version: 2026.7.5
chq-calendar-localstack  | Localstack returning with exit code 55. Reason:
chq-calendar-localstack  | License activation failed! 🔑❌
chq-calendar-localstack  | Reason: No credentials were found in the environment. Please make sure to
chq-calendar-localstack  | either set the LOCALSTACK_AUTH_TOKEN variable to a valid auth token.
```

```
$ docker compose ps -a
chq-calendar-localstack   exited   Exited (55)
```

**`./scripts/setup-local.sh` still exits 0.** Its readiness checks (`scripts/setup-local.sh:138-141`) cover ports 3000/3001/8000/8001 and never probe 4566, so a dead LocalStack is invisible.

Impact is contained — `backend/src/scripts/runArticleIngestLocalAws.ts` and `docs/runbooks/article-links.md` are the only consumers, and no automated test uses it — but the article-links local integration path is currently dead on a fresh clone.

**Suggested fix:** pin a known-good non-gated version instead of `:latest`, and add a `wait_for` entry for 4566 so the script reports the failure. An unpinned upstream `:latest` in a dev stack will drift again regardless.

---

## 2. DynamoDB tables are never created, and setup claims otherwise

`scripts/setup-local.sh:176` prints:

```
   • DynamoDB tables will be created automatically
```

Nothing creates them. On a fresh stack the API is broken:

```
$ curl -s -X POST -H 'Content-Type: application/json' -d '{}' http://localhost:3001/calendar
{"error":"Internal server error","message":"Cannot do operations on a non-existent table"}
# HTTP 500
```

Running `npm run init-tables -w backend` inside the container fixes it, after which the same call returns 200. Because DynamoDB Local runs with `-inMemory`, **this must be repeated after every container restart** — the data and the schema are both discarded.

### 2a. `init-tables` does not create the feedback table

Even the working path is incomplete. `backend/src/scripts/initTables.ts:166-167` creates exactly two tables:

```ts
await createEventsTable();
await createDataSourcesTable();
```

But `docker-compose.yml:33-35` declares three, and `backend/src/handlers/adminHandler.ts:140` reads `FEEDBACK_TABLE_NAME` (default `chautauqua-calendar-feedback`) at four call sites. Admin feedback endpoints therefore fail locally even after `init-tables` succeeds.

**Suggested fix:** run table init automatically from `setup-local.sh` (and ideally on backend container start, given `-inMemory`), and add the feedback table to `initTables.ts`. Failing that, correct the message — claiming automatic creation is worse than saying nothing.

---

## 3. `scripts/create-tables.sh` is unusable and creates the wrong table names

Two independent problems.

**Wrong names.** It creates `chq-calendar-*`, but the app reads `chautauqua-calendar-*`:

| `scripts/create-tables.sh` | `docker-compose.yml:33-35` |
|---|---|
| `chq-calendar-events` (:12) | `chautauqua-calendar-events` |
| `chq-calendar-data-sources` (:28) | `chautauqua-calendar-data-sources` |
| `chq-calendar-feedback` (:39) | `chautauqua-calendar-feedback` |

Even on success it produces three tables the application will never read, and leaves the "non-existent table" error from item 2 fully intact.

**Requires the AWS CLI.** It shells out to `aws dynamodb`, which is not a documented prerequisite and is not needed anywhere else in local development. Notably, **no test depends on the AWS CLI** — every AWS-touching test mocks the SDK — so this script is the only thing that would drag that dependency into a dev setup.

It is reachable as `npm run create-tables -w backend`.

**Suggested fix:** delete it and point to `init-tables` (the SDK-based script that uses correct names and needs no CLI). If it is kept for non-Docker use, fix the names and document the CLI prerequisite.

---

## 4. `npm run test:integration` points at a path that does not exist

`package.json:21`:

```json
"test:integration": "node integration-test.js",
```

The file lives at `utils/integration-test.js` (documented in `utils/README.md` as `node utils/integration-test.js`), so the npm script points at a path that does not exist and fails immediately with `MODULE_NOT_FOUND`. Presumably the file moved into `utils/` without the script being updated.

**Suggested fix:** point the script at `utils/integration-test.js`.

---

## Common thread

Items 1 and 2 are both cases of the setup script reporting success over a broken environment — exactly what the `wait_for` polling and the "fail loudly rather than printing the success banner" logic added in #214 were meant to prevent. The gap is that those checks cover only the four HTTP services, not LocalStack and not schema provisioning.

## Verified working (for contrast)

So the report is not read as "local dev is broken" — on the same clean checkout:

- Backend: 909 tests / 71 suites; all six Lambda handlers bundle
- Frontend: 1099 tests / 88 files; Vite build clean; **app renders correctly in-browser**, confirming the `@preact/preset-vite` blank-page regression from #214 is genuinely fixed
- iOS: 789 tests / 57 suites (Xcode 26.6, iOS 26.5 sim), widgets target builds under the app scheme
- publisher-format: 18 tests / 6 suites; workspace symlink resolves after a clean `npm ci`
- Terraform: `fmt -check -recursive` and `validate` both pass

The four core Docker services come up healthy and the frontend talks to the backend. The defects above are confined to LocalStack, schema provisioning, one dead script, and one stale path.

## Minor, not worth separate issues

- `ios/README.md:69,75` pins `OS=26.1` in its local `xcodebuild` recipe; that runtime will not exist on most machines. The surrounding text does say to swap it, so this is a papercut.
- `.github/workflows/ios.yml` comments cite a "716-test suite" in two places; it is now 789. That comment is load-bearing — it explains why `-parallel-testing-enabled NO` exists.
- `terraform validate` fails on a fresh clone with 8 `filebase64sha256(".../lambda-function.zip"): no such file` errors until `npm run package:terraform -w backend` has run. Not a config bug, but undocumented, and no CI job runs Terraform to catch it.
- There are two entry points to local setup and only one is documented. `package.json:20` defines `"setup": "./scripts/setup-local.sh"`, but no doc mentions `npm run setup` — README.md and docs/DEPLOYMENT.md invoke `./scripts/setup-local.sh` by path in all six places they reference it. The alias's only other mention is a code comment at `scripts/setup-local.sh:144`. Worth either documenting the alias or dropping it.

Woodwell:
## What changed

Four fixes for the local dev environment defects in #247: LocalStack pinned to a working Community release and added to setup's health checks; DynamoDB tables (including the previously missing feedback table) created automatically on every backend container start; the stale `scripts/create-tables.sh` deleted; `test:integration` pointed at the file's real path.

## Why

Fixes #247. All four share the failure mode #214 set out to eliminate: `./scripts/setup-local.sh` exited 0 over a broken environment. LocalStack was dead on arrival (`:latest` now resolves to the license-gated 2026.x line, exit 55), no script created the DynamoDB tables the backend reads (POST `/calendar` → 500 on a fresh stack), and two npm scripts were broken at invocation.

Design notes:

- **Tables are created on *every* backend container start** (`Dockerfile.dev` chains `npm run init-tables && npm run dev`), not just at first setup — DynamoDB Local runs `-inMemory`, so the schema vanishes on every restart. `initTables.ts` gains a bounded `waitForDynamoDB()` poll because `depends_on` orders container start, not readiness. Because init chains before the dev server, `setup-local.sh`'s existing backend health check now also implies the schema exists.
- **The feedback table is new to `initTables.ts`** — schema mirrored from `infrastructure/main.tf` (`id` hash key + `TimestampIndex` GSI). `adminHandler.ts` reads it at four call sites, so admin feedback endpoints failed locally even after a manual init.
- **LocalStack is pinned to exactly `4.14.0`**, the newest Community release, verified to boot unauthenticated and serve S3. If you'd rather track the major line (`localstack/localstack:4`) for patch updates, happy to change — exact pin was chosen for reproducibility.
- `create-tables.sh` was doubly broken: it created `chq-calendar-*` tables while the app reads `chautauqua-calendar-*`, and it required the AWS CLI, which nothing else in local dev needs (all AWS-touching tests mock the SDK).

## Verification

- [x] `npm run validate --workspace=backend` — clean (tsc + eslint `--max-warnings=0`)
- [x] `npm run test:ci --workspace=backend` — 909 tests / 71 suites pass; `src/scripts/*` is excluded from coverage so the floor is unaffected
- [x] `npm run build --workspace=frontend` — clean (1099 tests / 88 files + vite build; only root `package.json` touched on the frontend side)
- [x] Full `./scripts/setup-local.sh` from scratch, then verified independently of the script's own output:
  - 5/5 containers running, LocalStack **healthy** (was `Exited (55)`)
  - all three tables exist; `POST /calendar` → 200; `GET /admin/api/feedback` → `{"feedbacks":[]}` (was 500 "Cannot do operations on a non-existent table")
  - LocalStack S3 round-trip on 4566: create bucket / put / get all 200
- [x] **Restart recurrence** — the case first-run-only init would miss: `docker compose restart`, backend logs show tables being re-created (not "already exists"), `POST /calendar` → 200 with no manual step
- [x] With LocalStack deliberately broken, `setup-local.sh` now exits 1 and names the failing service (previously exited 0)

Not fixed here (listed as minor in #247): the `ios/README.md` OS pin, the stale test-count comment in `ios.yml`, the `terraform validate` build-order note, and the undocumented `npm run setup` alias.

Heads-up on merge: this touches root `package.json` and `backend/**`, which match both `FRONTEND_PATHS` and `BACKEND_PATHS` in `deploy-production.yml`, so both area deploys will fire. Both are content no-ops — no Lambda bundle or frontend build input changes — and the post-deploy smoke will run as usual.

## App Store listing

- [x] Not applicable — no visual change


bbernstein (approved):
lgtm. This is cleaner and ensures it is set up correctly each time.

PR #249, opened 2026-08-20
fix(e2e): pin the rail suite's clock so browser-checks stops going red at night

bbernstein:
`browser-checks` has been red on `main` every night and green every morning **on the same commit**: failing at 01:31Z, again at 01:58Z on a re-run, passing at 10:17Z. Deterministic, not flaky — and it will keep happening regardless of what anyone merges.

## The cause

Reproduced locally by pinning the browser clock to CI's exact failing instant. With the new diagnostic in place, `11c` now says:

```
today=2026-08-19 anchor=2026-08-27 mounted=2026-08-20,2026-08-21,2026-08-22 button=1
```

Today is not mounted at all. Once today's last event plus `Now`'s one-hour grace has passed, today leaves the `Now` window — so the anchor cannot return to it, and `⟳ Now` correctly stays visible. **The app is right; the check's assumption that today is always reachable is what was wrong.**

The boundary tracks the day's programming rather than the clock, which is why no fixed "fails after N o'clock" rule ever fit the history: pinned to `01:30Z` that night, `2026-08-19` is still mounted and `11c` passes; one minute later at `01:31Z` it is gone and `11c` fails. The earlier timezone pin (#241/#242) fixed the afternoon-**UTC** version of this and could not fix the late-evening-**Eastern** one.

## The fix

Pin the browser clock to mid-morning Institution time **on the run's own calendar day**.

- `setFixedTime`, not `install` — it pins what `Date.now()`/`new Date()` report while leaving every timer running. The app leans on real timers for the search debounce, the render window's observers and the rail's scroll settling; faking those would break the interactions these checks drive.
- The date stays the real today, so every assertion about which days carry events stays valid against live production data. Only the *hour* is removed.
- `14:00Z` is 10:00 EDT in season and 09:00 EST out of it — both comfortably mid-morning, so it buys determinism without offset arithmetic to get wrong twice a year.

## `11c` now explains itself

Every previous investigation into this check had to guess, because the failure line was `11c ⟳ Now hides once back on today:` with an empty detail. It now reports what the app thinks today is, where the anchor landed, and which days are mounted — so the next failure states its own cause.

## Verification

| | result |
|---|---|
| pinned mid-morning (the fix) | 36/36 |
| pinned to CI's failing instant | 35/36 — the exact CI failure, reproduced on demand |
| full `test:browser` suite | 36/36, 34/34, 16/16 |

The middle row is the falsification: the check can still fail, and fails for exactly the reason CI failed.

Found while merging #245, which touched **zero** frontend files — it only surfaced this by triggering a run after the boundary.

https://claude.ai/code/session_01Dxr9zVJLDHofJBEs26bUw4

### DIFF
--- a/frontend/e2e/fixedNow.mjs
+++ b/frontend/e2e/fixedNow.mjs
@@ -1,0 +1,35 @@
+/**
+ * One pinned instant, shared by the browser checks that care what time it is.
+ *
+ * These suites run against live production data, so "today" has to stay the
+ * real today — hardcode a date and every assertion about which days carry
+ * events goes stale the moment the feed moves on. What is removed is only the
+ * *hour*, which is the part that was making CI red at night and green in the
+ * morning on the same commit.
+ *
+ * `14:00Z` is 10:00 EDT in season and 09:00 EST out of it. Both are
+ * comfortably mid-morning — late enough that the day's programming is under
+ * way, early enough that it has not run out — which is all these checks need,
+ * and it buys that without offset arithmetic to get wrong twice a year.
+ *
+ * Lives in its own module so the suites cannot drift apart on what "now"
+ * means; two copies of this rule would eventually disagree.
+ */
+
+/** Mid-morning Institution time on the run's own calendar day. */
+export const FIXED_NOW = new Date(
+  `${new Intl.DateTimeFormat('en-CA', { timeZone: 'America/New_York' }).format(new Date())}T14:00:00Z`
+);
+
+/**
+ * Pins `page`'s clock. Call before `goto`.
+ *
+ * `setFixedTime`, not `install`: it pins what `Date.now()`/`new Date()` report
+ * while **leaving every timer running**. The app leans on real timers — the
+ * search debounce, the render window's observers, the rail's scroll settling —
+ * so faking those as well would break the very interactions these suites
+ * drive.
+ */
+export async function pinClock(page) {
+  await page.clock.setFixedTime(FIXED_NOW);
+}
--- a/frontend/e2e/verify-filter-reveal.mjs
+++ b/frontend/e2e/verify-filter-reveal.mjs
@@ -26,24 +26,21 @@ const check = (name, ok, detail) => {
 const browser = await chromium.launch();
 
 async function phone({ reducedMotion } = {}) {
-  // Chautauqua's own timezone, pinned rather than inherited from the runner.
+  // Chautauqua's own timezone, kept for belt-and-braces — it is not
+  // load-bearing. The paragraph that used to be here claimed the app "treats
+  // the browser's clock as event-time" and that whether `now` should be
+  // evaluated in the Institution's timezone was an open product question. Both
+  // were true when written and #243 ("resolve every date in the Institution's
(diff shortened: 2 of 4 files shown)


bbernstein:
## Review iteration complete — ready for merge

All review comments have been addressed. The latest Copilot pass on `7d5b1af` reports no new comments **and carries no `Suppressed comments` block** — which on this repo is the meaningful signal, since five earlier passes said "generated no new comments" while holding real findings in that collapsed section.

### Findings addressed

| Finding | Resolution | Commit |
|---|---|---|
| `11c` failed on `main` at night, passed in the morning, same commit | Clock pinned to mid-morning Institution time on the run's own calendar day; root cause is today's last event plus `Now`'s one-hour grace, not an hour of the clock | `e351869` |
| `11c` printed no detail, so every prior investigation had to guess | Reports `today`, `anchor`, `mounted`, `button` on failure | `e351869` |
| The file's timezone paragraph was stale post-#243 and had misled a reader into believing there was an unfixed product bug | Recorded as wrong rather than deleted, with the evidence (whole suite passes 36/36 under `Asia/Tokyo`) | `ab70d00` |
| `verify-timezone.mjs` tolerated clock skew by reading its baseline twice and accepting either | Clock pinned there too; checks 4 and 5 are now exact against a single baseline. Falsified: skewing one zone 36h fails checks 3/4/5 (11/16), restored 16/16 | `bab1892` |
| Grammar: "The two the pinned clock exists for" | Missing noun added | `50d4343` |
| `newPage()` leaked a `BrowserContext` per check | Context lifetime tied to the page; `verify-filter-reveal.mjs` had the same leak and got the same fix | `26ccd3f` |
| `.slice(0, 3)` truncated `11c`'s mounted-day diagnostic | Removed. Not hypothetical — the same passing run reports **nine** mounted days, so six were being hidden, and a `today` landing fourth would have been concealed in exactly the failure the diagnostic exists to explain | `7f050b5` |
| `verify-filter-reveal.mjs` carried a verbatim copy of the stale pre-#243 paragraph | Corrected; two stale copies would have re-set the same trap | `7d5b1af` |

The `.slice(0, 3)` finding was flagged four times before I acted on it, because I was reading review summaries and never opening the **inline** comments — a separate surface from the body's suppressed block. Worth recording as the process failure it was.

### Threads to resolve manually

All are addressed in the commits above; none needs further code change:

- `verify-rail.mjs` — Copilot and claude[bot] on `.slice(0, 3)` (fixed in `7f050b5`)
- `verify-rail.mjs` — Copilot and claude[bot] on the unclosed `BrowserContext` (fixed in `26ccd3f`)
- `verify-timezone.mjs` — claude[bot] grammar nit (fixed in `50d4343`)
- `verify-filter-reveal.mjs` — claude[bot] on the stale pre-#243 comment (fixed in `7d5b1af`)

### Checks

All 7 pass on `7d5b1af`: browser-checks, claude-review, screenshot-freshness, frontend ×2, backend ×2. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, no pending reviewers, not a draft.

Local verification on the merge tree: `36/36`, `34/34`, `16/16` across the three browser suites, and `npm run validate` clean.

**Ready for merge.** Not merging — that's the author's call.

PR #270, opened 2026-08-24
fix: an off-season landing, and browser checks that survive it (#269)

### LINKED ISSUE #269: browser-checks will go red from 2026-09-11: every suite waits on [data-day-key], which the off-season default screen never renders
## What will happen

From **2026-09-11 through 2026-09-30**, all three browser-check suites will fail on `main` — not from a code change, but from the calendar rolling past the season. Each suite gates its page bootstrap on a day section existing:

| File | Line | Wait |
|---|---|---|
| `frontend/e2e/verify-rail.mjs` | 77 | `waitForSelector('[data-day-key]', { timeout: 30000 })` |
| `frontend/e2e/verify-timezone.mjs` | 28 | same, 30 s |
| `frontend/e2e/verify-filter-reveal.mjs` | 55 | same, default timeout |

`data-day-key` (`DAY_SECTION_ATTR`, `lib/utils/daySections.ts:18`) is written only by a rendered day section (`EventListView.tsx:43`). When `filteredEvents.length === 0`, `page.tsx:620` renders `<EmptyState />` instead, so the attribute never appears and every `newPage()` hangs to its timeout. `verify-rail.mjs` alone calls `newPage()` **12 times**, so the job burns ~6 minutes doing nothing before failing.

## Why the default screen goes empty

Verified against the live feed and manifest (fetched 2026-08-24):

- `years.json` is `{"years":[2025,2026,2027],"defaultYear":2026}`, so the app defaults to **2026**, and `useEventData` scopes events to the selected year.
- The last 2026 event day in the feed is **2026-09-10**. The next event day of any kind is 2027-06-27 — a 290-day gap, and in a different year the default view cannot reach.
- The default scope for the current year is `next` (`useFilterState.ts:131`). Its window is `[now − 1 h, adaptiveEndDate)`. With no 2026 events left, `getAdaptiveEndDate` (`dateHelpers.ts:155-158`) takes its empty-input branch and returns `startDate + 91 days` — a window over an empty stretch of September–December 2026.
- Zero matches → `<EmptyState />` → no day section → every suite times out.

## Why it self-heals on 2026-10-01 (and why that is not a fix)

`generateYearsManifest` sets `defaultYear = now.getMonth() >= 9 ? year + 1 : year` (`eventsCalendarDataSyncService.ts:988`) — `getMonth()` is 0-indexed, so on **October 1** the manifest flips to `defaultYear: 2027`. The app then defaults to 2027, `getAdaptiveEndDate` scans forward to accumulate 50 events regardless of the gap, and the first 2027 events render — day sections come back and the suites should pass again.

So the outage is a **~20-day window, 2026-09-11 to 2026-09-30**, and it will recur every year in the same shape (the season's last event day → September 30). It is worth fixing rather than waiting out: three weeks of a red merge gate trains people to ignore it, and this is exactly the "boundary moves date to date" class that #249 and #268 each fixed one facet of (the hour, then the disabled-chip target).

*Correction to an earlier note of mine: I said the outage lasted until June 2027. The October manifest rollover means it does not — it is ~3 weeks.*

## Options

**1. Teach the harness the off-season regime (smallest).** Have the shared page bootstrap detect the empty default screen and either skip the day-section-dependent checks with an explicit "off-season, N checks skipped" line, or first switch to a year/scope that has events (`All Year`, or the year selector) before waiting. Risk: a skip that is silent, or that hides a real emptiness bug — the count must be printed, and "everything skipped" should still fail.

**2. Give the web app an off-season landing (largest, best product outcome).** iOS already has `OffSeasonLandingView`; the web has only `<EmptyState />`. A landing that says "the 2026 season has ended — browse the archive or preview 2027" would make the default screen useful for three weeks a year *and* give the harness something deterministic to assert. This is a real user-facing gap, not just a CI one: anyone visiting chqcal.org between Sep 11 and Oct 1 sees an empty list.

**3. Make `defaultYear` roll over when the season ends rather than on October 1.** Closes the gap at the source, but changes what every client defaults to and touches the sync service — the widest blast radius, and it would need thought about what "the season has ended" means for the archive.

Recommendation: **2 with 1 as the immediate stopgap**, since option 2 also removes the empty screen real visitors would see. Whatever ships, the guard should be proven by pinning the clock past 2026-09-10 and watching the suite pass for the right reason — `fixedNow.mjs` already supports a pinned instant, and #268's regime matrix (nine pinned dates) is the pattern to copy.

## Notes

- Discovered while fixing #268 (the season-tail disabled-chip failure) and deliberately left out of that PR's scope: it is a different regime with a different fix, and #268 was gating four merges.
- Nothing here is a regression from the 2026-08-24 merges; the same timeout reproduces on any commit.

bbernstein:
Closes #269.

## What was going to happen

From **2026-09-11 to 2026-09-30** all three browser-check suites would have gone red on `main` with no code change, and every visitor to chqcal.org would have seen "No events found — try adjusting your filters" with no filters set.

The last 2026 event day is 2026-09-10, the manifest still says `defaultYear: 2026`, and the current year's default scope is `next` — whose window falls back to `startDate + 91 days` on empty input. Zero matches means `EmptyState`, which writes no `data-day-key`, which every suite waits 30s for. `verify-rail` opens twelve pages, so the job burned ~6 minutes proving nothing and then failed with no diagnosis.

It self-heals on October 1 when the manifest rolls to 2027, and recurs every year in the same shape.

## What this does

**The web gets an off-season landing**, porting iOS's `LandingState` and `OffSeasonLandingView`. It names why the list is empty, counts down to the next opening, and offers the archive and a next-season preview. `EmptyState` is unchanged apart from a test id and still handles the case it was written for — the reader filtered to nothing.

**The browser checks became regime-aware.** `enterList` races the day section against the landing, and off-season taps a mid-season rail chip to widen the window into the season, so every geometry check still runs. Only two checks skip, both with printed reasons and no subject off-season. A suite where no check passed now exits non-zero.

**`verify-offseason.mjs`** runs a five-entry regime matrix on every push so the path cannot rot during the eleven months it is unreachable.

## Results

| | in season (prod) | off season (pinned 2026-09-15) |
|---|---|---|
| `verify-rail` | 37 / 0 / 0 | 32 / 0 / **2 skipped** |
| `verify-filter-reveal` | 35 / 0 / 0 | 34 / 0 / **1 skipped** |
| `verify-timezone` | 17 / 0 / 0 | 17 / 0 / 0 |
| `verify-offseason` | 14 / 0 / 0 | — (pins its own clock) |

Also green at the September 30 edge. Unit suite 1,174; backend 909.

## Five things the runs found that reading did not

- **Pinning the clock cannot reproduce the October 1 self-heal.** `defaultYear` is server-generated; `getDefaultYear()` is only the failed-fetch fallback. That matrix entry stubs `years.json`.
- **Pre-season is unreachable with a published feed.** `next`'s adaptive window keeps reaching forward until it has 50 events however far off, so the list is never empty and `in-season` is correct. The countdown belongs to a year announced in the manifest whose programme is not up yet — so that entry stubs the events feed empty.
- **My prediction about the archive-entry route was wrong.** I expected it to remove "Show earlier"; measured, it does not, because the 2026 feed carries events from January 3, outside the season. The rail chip stays primary for the reason that survives: a season whose events all fall inside it would have no earlier day.
- **Entering by chip scrolls, and a scrolled page is a different page** — past the sentinel the filter card is `inert`, so its buttons leave the accessibility tree. Check 1 still saw "All Season" (raw `textContent`); check 2's `getByRole` for the same button timed out. And **programmatic scrolling cannot undo a rail hold**: `scrollTo(0, 3100)` came back as `scrollY 523`. `enterList` settles with a wheel.
- **`verify-filter-reveal` check 10 picked a filter with no off-season subject.** "Today" matches nothing then: the list empties, the document collapses 11,042 → 985px, and the Filters toggle goes with it, so 10/11/15 all failed on a correct app.

## Verification

Falsified rather than assumed — each guard broken in turn and confirmed red, then green again:

| injected defect | caught by |
|---|---|
| landing branch disabled | `enterList` names the empty state in seconds; matrix 4 passed / 10 failed |
| archive entry disabled | `verify-rail` exits 1 rather than measuring an empty page |
| everything skipped | `finish()` exits 1 |
| `now < start` → `false` | matrix entry 3, plus 3 unit tests |

Two checks were rewritten *because* falsification exposed them: `3d` was passing on a page with no landing on it, and `1d`'s unconditional click aborted the suite instead of reporting entries 2–5. The DST unit test's "same day, same count" assertion also turned out not to discriminate — the naive implementation gave 248 for both hours — so the test now says which line is the guard.

`npx vite build` was used for the injections, since `npm run build` gates the bundle on the unit suite and would have served the previous bundle.

https://claude.ai/code/session_01CZUXs5vAgdfPGRx8GtCRAn

### DIFF
--- a/docs/superpowers/plans/2026-08-24-off-season-landing-269.md
+++ b/docs/superpowers/plans/2026-08-24-off-season-landing-269.md
@@ -1,0 +1,2125 @@
+# Off-Season Landing and Regime-Aware Browser Checks Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Give the web app an off-season landing screen, and make the three
+Playwright browser-check suites work in the off-season regime, so that neither
+visitors nor CI hit a blank screen between the season's last event day and the
+years-manifest rollover on October 1.
+
+**Architecture:** A pure `determineLandingState` function ports iOS's
+`LandingState` to TypeScript; a new `OffSeasonLanding` component replaces
+`<EmptyState />` in `page.tsx` when the reader has narrowed nothing and the
+season is genuinely over or not yet begun. On the harness side, the three
+suites' copy-pasted bootstrap and tally collapse into two shared modules — one
+that detects the regime and navigates into the season when the default screen
+is the landing, and one that reports pass/fail/skip honestly. A new
+`verify-offseason.mjs` runs a five-entry regime matrix on every CI push so the
+off-season path cannot rot during the eleven months it is unreachable.
+
+**Tech Stack:** Vite 7, Preact 10 (via `preact/compat`, imports from
+`'react'`), TypeScript 5, Tailwind CSS 4, Vitest + `@testing-library/preact`,
+Playwright (Chromium only), Node 24.
+
+**Spec:** `docs/superpowers/specs/2026-08-24-off-season-landing-269-design.md`
+
+**Issue:** [#269](https://github.com/bbernstein/chq-calendar/issues/269)
+
+## Global Constraints
+
+- **Never commit to `main`.** Work happens on `fix/269-off-season-landing`,
+  which already exists and already holds the spec commit.
+- **Preact, not React.** Import hooks and React-shaped types from `'react'` —
+  `@preact/preset-vite` aliases it to `preact/compat` at build time and
+  `vitest.config.ts` does the same. This is the project convention; see
+  `CLAUDE.md`. Files that render JSX and wire DOM handlers must keep the
+  `'react'` import.
+- **`@/` path alias** maps to `frontend/src/`.
+- **Institution time, always.** Never use `new Date(y, m, d)`, `getMonth()`,
+  `getFullYear()`, or millisecond date arithmetic for calendar reasoning. Use
+  `chqParts` / `chqDateAt` / `dayKeyOf` from `@/lib/utils/chqTime` and
+  `@/lib/utils/dayWindow`. Adding `86_400_000` ms across a DST transition
+  lands an hour out and produces the wrong day.
+- **Tailwind CSS 4**, dark mode via `prefers-color-scheme` (so every color
+  class needs its `dark:` counterpart), mobile-first with `sm:` / `lg:`.
+- **Verification before every commit**, from `frontend/`:
+  `npm run build` (runs `validate` — type-check + lint — then the unit suite,
+  then `vite build`).
+- **Coverage floor** is enforced by `.coverage-floor.json`; new modules need
+  real tests or the build fails.
+- **Browser-check suites are ESM `.mjs` run directly by Node**, not by a test
+  runner. They print `PASS`/`FAIL` per check and exit non-zero at the end.
+  They run against **live production data** — CI builds the branch, serves it
+  with `vite preview`, and `vite.config.ts` proxies `/cache` to
+  `https://www.chqcal.org`. Never write an assertion that hardcodes a
+  calendar date; derive it from the feed.
+
+---
+
+## File Structure
+
+**New:**
+
+| File | Responsibility |
+|---|---|
+| `frontend/src/lib/utils/landingState.ts` | Pure `determineLandingState`; no clock read, no DOM |
+| `frontend/src/__tests__/lib/utils/landingState.test.ts` | Its unit tests (util tests live under `src/__tests__/lib/utils/`, not beside the source) |
+| `frontend/src/components/layout/OffSeasonLanding.tsx` | The landing screen; presentational, all state passed in |
+| `frontend/src/components/layout/__tests__/OffSeasonLanding.test.tsx` | Its component tests |
+| `frontend/e2e/results.mjs` | Shared `check` / `skip` / `finish` tally |
+| `frontend/e2e/regime.mjs` | Shared `enterList` bootstrap + regime detection |
+| `frontend/e2e/verify-offseason.mjs` | The pinned-clock regime matrix |
+
+**Modified:**
+
+| File | Change |
+|---|---|
+| `frontend/src/app/page.tsx` | Derive `landingState`; branch to `OffSeasonLanding` at the empty case |
+| `frontend/src/components/layout/EmptyState.tsx` | Add `data-testid="empty-state"` (nothing else) |
+| `frontend/e2e/fixedNow.mjs` | `E2E_NOW` override |
+| `frontend/e2e/verify-rail.mjs` | Use `enterList` + shared tally; skip checks 3 and 11 off-season |
+| `frontend/e2e/verify-timezone.mjs` | Use `enterList` + shared tally |
+| `frontend/e2e/verify-filter-reveal.mjs` | Use `enterList` + shared tally |
+| `frontend/e2e/README.md` | Document the regime bootstrap, `E2E_NOW`, and the new suite |
+| `frontend/package.json` | Add `verify-offseason.mjs` to `test:browser` |
+
+---
+
+## Task 1: `determineLandingState`
+
+**Files:**
+- Create: `frontend/src/lib/utils/landingState.ts`
+- Test: `frontend/src/__tests__/lib/utils/landingState.test.ts`
+
+**Interfaces:**
+- Consumes: `getChautauquaSeasonWeeks` from `@/lib/utils/dateHelpers`,
+  `dayKeyOf` from `@/lib/utils/dayWindow`.
+- Produces:
+  - `export type LandingState` — the discriminated union below.
+  - `export function determineLandingState(input: LandingStateInput): LandingState`
+  - `export interface LandingStateInput { now: Date; selectedYear: number; availableYears: number[]; yearHasEvents: boolean }`
+
+Task 2 imports the type; Task 3 calls the function.
+
+### Background the implementer needs
+
+The Chautauqua season is nine weeks starting at **noon on the Saturday before
+the 4th Sunday of June**. `getChautauquaSeasonWeeks(year)[0].start` is exactly
+that instant, and it matches iOS's `SeasonCalendar.seasonStart(year:)` — do
+not reimplement it.
+
+Day counts must be whole *Institution calendar* days. The safe way is to
+reduce both instants to their `yyyy-mm-dd` day key (which `dayKeyOf` resolves
+in `America/New_York`) and subtract those as pure UTC calendar dates. Pure
+calendar dates have no DST in them, so the subtraction is exact.
+
+- [ ] **Step 1: Write the failing test**
+
+Create `frontend/src/__tests__/lib/utils/landingState.test.ts`:
+
+```ts
+import { describe, it, expect } from 'vitest';
+import { determineLandingState } from '@/lib/utils/landingState';
+import { getChautauquaSeasonWeeks } from '@/lib/utils/dateHelpers';
+import { chqDateAt } from '@/lib/utils/chqTime';
+
+const opening = (year: number) => getChautauquaSeasonWeeks(year)[0].start;
+
+describe('determineLandingState', () => {
+  it('is pre-season before the selected year opens', () => {
+    const state = determineLandingState({
+      now: chqDateAt(2026, 3, 1, 10),
+      selectedYear: 2026,
+      availableYears: [2025, 2026, 2027],
+      yearHasEvents: true,
+    });
+    expect(state.kind).toBe('pre-season');
+    if (state.kind !== 'pre-season') return;
+    expect(state.opening.getTime()).toBe(opening(2026).getTime());
+    expect(state.daysUntil).toBeGreaterThan(100);
+  });
+
+  it('is post-season once the season has opened and the year has events', () => {
+    const state = determineLandingState({
+      now: chqDateAt(2026, 9, 15, 10),
+      selectedYear: 2026,
+      availableYears: [2025, 2026, 2027],
+      yearHasEvents: true,
+    });
+    expect(state).toEqual({
+      kind: 'post-season',
+      endedSeasonYear: 2026,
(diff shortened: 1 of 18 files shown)


bbernstein:
Review finding addressed in `9ae1b6c` — it was correct, and the failure mode it described is one I had not considered.

Entry 5 stubbed `years.json` to roll `defaultYear` forward but left `all-events-<nextYear>.json` pointing at production, so `5a` was effectively asserting that next year's season is already published on chqcal.org. True today; a fact about production data rather than about this code.

The scenario that matters is the one the review named: in the very window #269 exists to fix, a year can roll over in the manifest before its programme goes up. Production would then correctly show "Almost showtime" and `5a` would go red on every PR merged in those weeks, for a reason unrelated to the change under review — and on the default CI path, since `test:browser` has no path filtering.

**Fix:** next year's feed is now a three-event mid-July fixture (mid-July is inside the season in every year). `5a` reports `days=3` — the fixture, not production's 2027 programme. `emptyEvents` generalised to `routeEvents: { glob, body }` so the pre-season entry and this one share one mechanism.

**Falsified against the review's exact scenario.** With the stub set to `{ data: [] }`:

```
FAIL  5a rolling the manifest forward ends the empty screen
      — day=2026-10-05 defaultYear=2027 days=0 landing=true heading=Almost showtime
```

That reproduces the predicted false-red precisely, and it is now unreachable from production data.

Re-verified after the change: `test:browser` 37 / 35 / 17 / 14, zero failures; unit suite 1,174; lint clean.

bbernstein:
All three inline findings addressed in `682db1c`. All were real; one needed a different fix than the one suggested.

**1. `OffSeasonLanding.tsx:71` — redundant `as number`.** The concern is right, but the prescribed fix does not compile. Both reviewers suggested calling `onPreviewNextSeason(state.nextSeasonYear)` directly since TS narrows it inside the `!== null` guard. Narrowing a *property access* does not survive into a closure — TS cannot know the property is unchanged by the time the handler runs:

```
src/components/layout/OffSeasonLanding.tsx(71,50): error TS2345:
  Argument of type 'number | null' is not assignable to parameter of type 'number'.
```

Hoisted to a local `const` instead, which narrows properly (a `const` cannot be reassigned) and needs no cast. The comment records why, so the cast does not come back.

**2. `results.mjs:14` — truthiness drops a falsy detail.** Fixed to `detail != null`. Falsified:

```
PASS  numeric zero detail — 0        (previously printed nothing after the name)
PASS  empty-string detail —
PASS  undefined detail
```

**3. `verify-offseason.mjs:106` — no short-circuit after check 0.** It carried an undefined edge into `shift()`, which threw and aborted with a stack trace instead of a PASS/FAIL summary. Now stops there. Falsified by stubbing the edges undefined:

```
FAIL  0 feed has a season to reason about — undefined … undefined
0 passed, 1 failed, 0 skipped
No check passed — nothing was proved.
```

Still exits non-zero, as before — the gain is that a temporarily unavailable feed reads as an actionable failure rather than a crash.

Re-verified after all three: `test:browser` 37 / 35 / 17 / 14, unit suite 1,174, type-check and lint clean.

bbernstein:

Four findings across two review passes, all real, all fixed and falsified:

| finding | fix | falsified by |
|---|---|---|
| entry 5 depended on live production data for next year's events | hermetic three-event fixture | stub set to `{data: []}` → reproduced the predicted false-red exactly |
| `as number` cast could mask a type regression | hoisted to a local `const` (see note below) | `tsc --noEmit` |
| `check()` dropped a falsy `detail` | `detail != null` | `check('x', true, 0)` now prints `— 0` |
| no short-circuit after check 0 | stops with a clean summary | edges stubbed undefined → `No check passed — nothing was proved` |

**One note, since the suggested fix was itself wrong.** Both passes proposed calling `onPreviewNextSeason(state.nextSeasonYear)` directly, on the reasoning that TS narrows it inside the `!== null` guard. It does not compile — narrowing a *property access* does not survive into a closure, because TS cannot prove the property is unchanged when the handler runs (`error TS2345: Argument of type 'number | null'`). Hoisting to a local `const` narrows correctly, since a `const` cannot be reassigned. The concern was right; the prescription wasn't.

### Threads to mark resolved

All five are already **outdated** — the code each points at has changed — so none block the merge. Listing them for manual resolution:

- `frontend/e2e/verify-offseason.mjs:106` (Copilot) — short-circuit after check 0
- `frontend/src/components/layout/OffSeasonLanding.tsx` (Copilot) — redundant cast
- `frontend/e2e/results.mjs:23` (Copilot) — truthy `detail`
- `frontend/src/components/layout/OffSeasonLanding.tsx` (claude) — redundant cast
- `frontend/e2e/results.mjs` (claude) — truthy `detail`

### Checks

All seven pass: `test-frontend` (24, 25), `test-backend` (24, 25), `browser-checks`, `screenshot-freshness`, `claude-review`. The three `skipping` entries are the documented same-repo `push`/`pull_request` dual-listing — see the "In CI" section of `frontend/e2e/README.md`.

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, no pending reviewers.

Locally re-verified at `682db1c`: `test:browser` 37 / 35 / 17 / 14 in season, and pinned to 2026-09-15 the three regime-aware suites give 32+2 skipped / 34+1 skipped / 17 — zero failures in either regime, plus the September 30 edge.

**This PR is ready for merge.** Leaving the merge itself to you.

bbernstein:
Addressed the outstanding Copilot finding in `7ce158d`.

**It was in the collapsed `Suppressed comments` block**, not in the visible summary — which on this repo is where real findings have hidden before while the summary said "0 new comments". Worth flagging for anyone reading this thread later: `gh pr view --json reviews` does not surface bot reviews at all, so this needed the GraphQL `reviews` node to find. Copilot has reviewed this PR three times; only the GraphQL query showed it.

### The finding

`verify-filter-reveal.mjs:139` used `?.click()` on a `find(...)` result. If the button is missing or renamed the click is a silent no-op — the panel then stays open *because nothing happened*, and check 10 passes having proved nothing about dismissal. A vacuous pass sitting in the middle of a PR whose whole subject is vacuous passes.

The line was mine: I changed the target from `Today` to `All Season` one commit earlier, which is exactly the edit that can silently stop matching.

**Fix:** the `evaluate` now reports whether the button was found, and check 10 is conjoined with it. Falsified by renaming the target to one that cannot match:

```
FAIL  10 a filter change does NOT dismiss
      — NO All Season button found — the click was a no-op
```

Before this change, that scenario printed `PASS`.

**Swept the other sites:** `verify-rail.mjs:212` uses a bare `.click()` (throws rather than no-ops) and is already guarded by a preceding count check; `pickRef`'s `find` is a read, not a click. No other instances.

### The second suppressed comment

`results.mjs` truthy-`detail` — already fixed in `682db1c`; Copilot's quoted context shows the corrected code, so it was re-anchored stale rather than newly reported.

### Copilot's third review

Its pass on `682db1c` returned:

> Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.

So the current head has not had a Copilot pass. My attempts to re-request via both REST and the correct bot login did not register, presumably for the same quota reason. Flagging rather than papering over it: **if you want a Copilot review of the final state, it likely needs your quota.** `claude-review` has passed on every commit including the head.

### Verification

Re-verified both regimes after the fix — in season 37 / 35 / 17 / 14; pinned to 2026-09-15, 32 + 2 skipped / 34 + 1 skipped / 17. Zero failures either way. Unit suite 1,174, lint clean.

bbernstein:

Five findings across four review rounds (Copilot ×2, `claude-review` ×2), all real, each falsified against a reproduction before and after the fix:

| finding | fix | falsified by |
|---|---|---|
| matrix entry 5 depended on live production data for next year's feed | hermetic 3-event fixture | stub emptied → reproduced the predicted false-red |
| `as number` cast could mask a type regression | hoisted to a local `const` | `tsc --noEmit` |
| `check()` dropped a falsy `detail` | `detail != null` | `check('x', true, 0)` now prints `— 0` |
| no short-circuit after check 0 | clean summary instead of a stack trace | edges stubbed undefined |
| **`?.click()` no-op let check 10 pass vacuously** | click reports whether the button was found; check 10 conjoined with it | target renamed → `NO All Season button found — the click was a no-op`, where it previously printed `PASS` |

The last one was in Copilot's collapsed `Suppressed comments` block, and it was on a line I had changed one commit earlier. Two of the fixes also required a different remedy than the one suggested — the `as number` cast could not simply be deleted (property-access narrowing does not survive into a closure), and the earlier `docH`-threshold approach to check 13 stood the check down in season on a 53px reflow that had produced a perfect result.

### Threads to mark resolved

All five are **outdated** — the code each points at has changed — so none block the merge. Listing for manual resolution:

- `frontend/e2e/verify-offseason.mjs` (Copilot) — short-circuit after check 0
- `frontend/e2e/results.mjs` ×2 (Copilot, claude) — truthy `detail`
- `frontend/src/components/layout/OffSeasonLanding.tsx` ×2 (Copilot, claude) — redundant cast

### Checks

All seven pass on `7ce158d`: `test-frontend` (24, 25), `test-backend` (24, 25), `browser-checks`, `screenshot-freshness`, `claude-review`. The three `skipping` entries are the documented same-repo `push`/`pull_request` dual-listing — see the "In CI" section of `frontend/e2e/README.md`.

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, no pending reviewers, not a draft, zero commits behind `main`.

### Local verification at HEAD

- In season, against the branch preview: `verify-rail` 37, `verify-filter-reveal` 35, `verify-timezone` 17, `verify-offseason` 14 — zero failures.
- Pinned to `E2E_NOW=2026-09-15`: 32 + 2 skipped, 34 + 1 skipped, 17 — zero failures. Also green at the September 30 edge.
- Unit suite 1,174; backend 909; type-check and lint clean.

### One caveat, stated rather than papered over

Copilot's pass on the current head returned *"unable to review … has reached their quota limit"*, and re-requesting did not register. **The final state has had no Copilot review** — `claude-review` has passed on every commit including this one, and its latest pass independently re-verified all five fixes and found no new issues. If you want a Copilot pass before merging, it likely needs your quota.

**This PR is ready for merge.** Leaving the merge itself to you.

PR #271, opened 2026-08-24
test: filterHeader was exercising a page with no events in it

bbernstein:
Found while working on #269 ([#270](https://github.com/bbernstein/chq-calendar/pull/270)), deliberately left out of that PR's scope. Independent of it — different file, no overlap.

## The defect

`Event.categories` is `Array<{ name: string }>`, but `filterHeader.test.tsx`'s fixture used a bare string array:

```ts
categories: ['Lecture'],
```

`useEventData` maps every raw event through `decodeEventHtmlEntities` before `setEvents` is reached. That reads `cat.name` — `undefined` on a string — and then lowercases it while building `_tagsLowerSet`, throwing at `eventHelpers.ts:63`. The outer `catch` only logs, so `events` stays empty and the page renders as though the feed were down.

Throw site confirmed by stack frame rather than by reading:

```
message="Cannot read properties of undefined (reading 'toLowerCase')"
firstAppFrame=at map (src/lib/utils/eventHelpers.ts:63:56)
```

(An earlier version of this description, and of the comment in the test file, blamed `useEventData`'s own `normalizeTag`. That code never runs on this path — corrected in `840a5d8`.)

Measured both shapes against `page.tsx`:

| fixture | events loaded | parse errors | `Events (…)` readout |
|---|---|---|---|
| `['Lecture']` | 0 | 1 (swallowed) | `Events (0/0)` |
| `[{ name: 'Lecture' }]` | 1 | 0 | `Events (0/1)` |

## Why nothing noticed

Every assertion in the file is about header geometry — the sticky card, the sentinel, the parked and `inert` states — and all of it renders with or without events. The file has been green for its whole life while exercising a page holding zero of the two events it defines.

`renderPage`'s own comment said the rail *"only exists once events have loaded"*. It never did: the rail is built from the season's navigable bounds and renders regardless. That wording is part of why this went unnoticed, so it's corrected here too.

## The fix

The fixture is corrected, **and** a guard is added — swapping the fixture alone would leave the same door open for the next person. It asserts the app's own count from `ActiveFilters`:

```ts
expect(screen.getByText(/^Events \(\d+\/2\)$/)).toBeInTheDocument()
```

The denominator is pinned at 2; the numerator deliberately is not. How many events fall inside the default scope depends on when the suite runs — how many were *loaded* does not.

**Falsified:** reverting only the fixture, with the guard left in place, fails exactly that one test and nothing else.

## Scope

Swept the rest of the suite — `EventCard.titleDetails.test.tsx` already uses the object shape, and no other test feeds `useEventData`. This was the only offender.

Full suite: 1,151 tests, 89 files, build and lint clean.

**Not fixed here, worth its own decision:** `useEventData` swallows a malformed-feed parse failure into a `console.error` and renders an empty calendar. In production that means a bad feed looks like an empty season rather than an error. #270's `determineLandingState` rule 3 already ensures such a visitor gets the honest generic empty state rather than "See you next season", but the swallowing itself is untouched.

https://claude.ai/code/session_01CZUXs5vAgdfPGRx8GtCRAn

### DIFF
--- a/frontend/src/__tests__/integration/filterHeader.test.tsx
+++ b/frontend/src/__tests__/integration/filterHeader.test.tsx
@@ -138,13 +138,22 @@ describe('page.tsx — the sticky filter header', () => {
   // Asking for the right file is not the same as the app being able to read
   // it, and the difference is invisible from everything else in this suite.
   //
-  // `useEventData` derives categories via `event.categories.map(c => c.name)`.
-  // A fixture using a bare string array yields `undefined`, `normalizeTag`
-  // then calls `.toLowerCase()` on it, and the whole parse throws into a
-  // `catch` that only logs — so `events` stays empty and the page renders as
-  // though the feed were down. Every other assertion in this file still
-  // passed, because they are all about header geometry, which renders with or
-  // without events. This file spent its life exercising a page that had none.
+  // A fixture using a bare string array for `categories` throws before the
+  // fetch result is ever stored. `useEventData` maps every raw event through
+  // `decodeEventHtmlEntities` first (`useEventData.ts`, before `setEvents`),
+  // and that reads `cat.name` — `undefined` on a string — then calls
+  // `.toLowerCase()` on it while building `_tagsLowerSet`
+  // (`eventHelpers.ts:63`, confirmed as the throw site by stack frame, not by
+  // reading). The outer `catch` only logs, so `events` stays empty and the
+  // page renders as though the feed were down.
+  //
+  // Note it is NOT `useEventData`'s own category/tag derivation that chokes —
+  // that code never runs on this path. An earlier version of this comment
+  // said it did, which would have sent the next reader to the wrong function.
+  //
+  // Every other assertion in this file still passed throughout, because they
+  // are all about header geometry, which renders with or without events. This
+  // file spent its life exercising a page that had none.
   //
   // The count comes from `ActiveFilters`, so it is the app's own reading of
   // how many events it holds. The denominator is pinned and the numerator is
--- a/frontend/src/__tests__/integration/filterHeader.test.tsx
+++ b/frontend/src/__tests__/integration/filterHeader.test.tsx
@@ -47,7 +47,9 @@ function eventsPayload() {
         endDate: `${YEAR}-07-06T11:45:00`,
         location: 'Amphitheater',
         description: 'A lecture.',
-        categories: ['Lecture'],
+        // `Array<{ name: string }>`, per `Event` in lib/types — NOT a bare
+        // string array. See the parse guard below for what that costs.
+        categories: [{ name: 'Lecture' }],
       },
       {
         id: 'e2',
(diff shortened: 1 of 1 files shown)


bbernstein:
Finding addressed in `840a5d8`. It was correct, and I verified it by capturing the stack frame rather than by re-reading the code:

```
message="Cannot read properties of undefined (reading 'toLowerCase')"
firstAppFrame=at map (src/lib/utils/eventHelpers.ts:63:56)
```

So the throw is in `decodeEventHtmlEntities`, which `useEventData` applies to every raw event *before* `setEvents` is reached — it reads `cat.name` (`undefined` on a string) and then lowercases it while building `_tagsLowerSet`. `useEventData`'s own category/tag derivation, which my comment blamed, never runs on this path at all.

The fix and the guard were unaffected. The explanation was not, and it is checked into the test file where the next reader will trust it — which is exactly why it was worth correcting. The comment now names `eventHelpers.ts:63`, says the attribution was confirmed by stack frame rather than by reading, and explicitly notes that `useEventData`'s own derivation is *not* what chokes, so the correction does not get quietly undone later.

I have also corrected the PR description above.

Re-verified: 12/12 in that file, 1,151 across the suite, lint clean.

On the sandbox note — the `1,151 tests / 89 files` figure is from a local `npm run build` (which runs `vitest run --coverage` before `vite build`) on this branch, and `test-frontend` has since passed on both Node 24 and 25 in CI.

bbernstein:

One finding across two passes, and it was correct: the explanation comment blamed `useEventData`'s own `normalizeTag`, when the throw is actually in `decodeEventHtmlEntities` — which runs on every raw event *before* `setEvents`, so `useEventData`'s category derivation never executes on that path. Verified by stack frame rather than by re-reading:

```
message="Cannot read properties of undefined (reading 'toLowerCase')"
firstAppFrame=at map (src/lib/utils/eventHelpers.ts:63:56)
```

Corrected in `840a5d8`, in both the test-file comment and the PR description. The latest review re-traced the path independently and confirmed the new attribution matches the source.

### Threads to mark resolved

None outstanding — the finding came in the review body rather than as an inline thread, so there is nothing to resolve here.

### Checks

All seven pass: `test-frontend` (24, 25), `test-backend` (24, 25), `browser-checks`, `screenshot-freshness`, `claude-review`. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, no pending reviewers.

**This PR is ready for merge.** Leaving the merge itself to you.

Note on ordering: independent of #270 — different files, no conflict — so either can go first.

bbernstein:
Correction to my closing comment above, in the interest of not overstating the review coverage.

I said there were no outstanding threads. That is true, but it is true for a reason I had not checked: **Copilot never reviewed this PR at all.** GraphQL reports `reviews.totalCount: 0` and `reviewThreads.totalCount: 0`. On #270 Copilot hit a quota limit (`unable to review … has reached their quota limit`), and that most likely applies here too.

This matters because `gh pr view --json reviews` returns empty for bot reviews whether they exist or not — on #270 it showed nothing while GraphQL showed three Copilot reviews, one of them carrying a real finding in a collapsed `Suppressed comments` block. So "no findings" from that command means nothing either way.

What this PR has actually had: `claude-review` on both commits (it posts as an issue comment rather than a formal review), which produced one finding — the misattributed throw site — since fixed and independently re-verified by its follow-up pass.

So: all checks green, no outstanding comments, but **no Copilot pass on this change**. If you want one before merging, it likely needs your quota.

